PR #14, opened 2018-10-26
Set up return block in initializer

skunkworks:
## What

Sets up the return block handler in the PopupBridge initializer instead of in the script handler for the `popupBridge.open()` call.

## Why

To use PopupBridge as a general purpose messaging tool to pass arbitrary data from native to webview via `onComplete()`.

PopupBridge previously set up the return block handler after the `popupBridge.open()` call. By moving the setup of the return block into the initializer, we can call `onComplete()` to pass data into the webview without having to to call `open()` first.

This does not have any side effect on the primary use case of PopupBridge, which is to handle popups in webviews; it still behaves exactly the same.

## Concerns

What this PR does is enable a more general use case where an app needs to pass data into the webview. It's a useful thing to be able to do. I could see an argument for adding a separate method for this -- e.g.. open() and onComplete() and onMessageReceived().


crookedneighbor:
@skunkworks Can you add info to the README about how to use this?

PR #16, opened 2019-08-16
Remove extra header file POPPopUpBridge.h

rzulkoski:
It looks like POPPopUpBridge.h wasn't renamed/removed properly and there is still an extra file hanging out at PopupBridge/Classes/POPPopUpBridge.h as well as some confusion in the project file itself. It would appear that fixing up this header fixes an issue with bringing in the framework through Carthage.


billwerges:
Thanks for fixing this @rzulkoski

PR #18, opened 2019-10-29
Add isIosWebview property to popupBridge

skunkworks:
By having the iOS PopupBridge library set up the `popupBridge` JS object with this property, it makes it trivial for the web app's JS to detect when the web page has been loaded in an iOS webview.

### DIFF
--- a/Example/POPViewController.m
+++ b/Example/POPViewController.m
@@ -31,4 +31,8 @@
     [viewController dismissViewControllerAnimated:YES completion:nil];
 }
 
+- (void)popupBridge:(POPPopupBridge *)bridge receivedMessage:(NSString *)messageName data:(NSString *)data {
+    NSLog(@"Received message %@ with data %@", messageName, data);
+}
+
 @end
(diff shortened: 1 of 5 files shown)


crookedneighbor:
@skunkworks can you provide the use case for this?

skunkworks:
@crookedneighbor For Venmo webview integrations, they need to set up the `deepLinkReturnUrl` property based on whether their checkout page is being loaded in an iOS webview or an Android webview (or not in a webview at all). This is fairly onerous for merchants, who might be able to accomplish this by setting a custom user agent string for their webviews, or otherwise use user agent strings. It seems sensible to instead have PopupBridge clearly identify which platform the web app is being loaded from.

crookedneighbor:
@skunkworks they can't inject their own JS into the webview and use that to determine it? I feel like merchants should not be checking for the popupBridge object on the window in their own integrations.

skunkworks:
I think that's fair. Closing

crookedneighbor:
In fact, if they _did_ check for the popupBridge object themselves, that would also tell them that they are in a webview. No need for extra properties.

PR #25, opened 2020-05-06
Fix bug with returning to the app a second time

sestevens:
Fix a bug where returning from `SFSafariViewController` only worked the first time by not setting the returnBlock to nil.


PopupBridge/Classes/POPPopupBridge.m
```diff
@@ -55,7 +55,7 @@ - (id)initWithWebView:(WKWebView *)webView delegate:(id<POPPopupBridgeDelegate>)
 
                 NSMutableDictionary *payloadDictionary = [NSMutableDictionary new];
                 payloadDictionary[@"path"] = path;
-                payloadDictionary[@"queryItems"] = [self.class dictionaryForQueryString:url.query];
+                payloadDictionary[@"queryItems"] = [POPPopupBridge dictionaryForQueryString:url.query];
```
billwerges, line 58: These instances of `self.class` should be changed to `weakSelf.class`. In general we want to avoid using hard coded class names because it can break sub-classing by preventing method overriding.

sestevens: We figured it was OK to use the class name in this case since these are private methods that can't be overridden in a subclass. But I'm happy to change it to `weakSelf.class` if you prefer.

sestevens: Updated ✔️

scannillo:
Looks good + CHANGELOG entry

sestevens:
> Looks good + CHANGELOG entry

Good catch! I added a changelog entry. I also added a PR template to #24 so that we'll have a reminder about the changelog in the future 🙂

PR #34, opened 2021-07-26
Add GitHub Actions CI

scannillo:
### Summary of changes
 - Remove Travis CI
 - Replace with GH Actions
 - Add jobs for Xcode 13 beta builds
 - Update SPM demo app to use HTTPS URL & master branch as default
 
 ### To Do Next
 - There is no Carthage sample app. Either we add one, or at least make sure `carthage update` succeeds somewhere in CI.
 - We aren't accurately testing the Venmo flow since we can't on a simulator. Let's add a MockVenmo app ([similar to this PR](https://github.com/braintree/braintree_ios/pull/646)) so we can catch potential app-switch issues.

 ### Checklist

 - ~Added a changelog entry~

### Authors
@scannillo


.github/workflows/tests.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: actions/checkout@v2
+      - name: Use Xcode 12.5
+        run: sudo xcode-select -switch /Applications/Xcode_12.5.app
+      - name: Install CocoaPod dependencies
+        run: pod install
+      - name: Run UI Tests
+        run: xcodebuild -workspace 'PopupBridge.xcworkspace' -sdk 'iphonesimulator' -configuration 'Release' -scheme 'UITests' -destination 'name=iPhone 12,platform=iOS Simulator' test #TODO: add some sort of xcpretty or xcbeautify
+
+  # Xcode 13-beta Jobs
+  ui_test_job_xcode_beta:
```
scannillo, line 31: I only added the UI tests for XCode 13, in addition to 12.5, since I don't think it's that valuable to unit test both XCode versions.

sarahkoop (approved):

sshropshire (approved):

PR #45, opened 2023-03-10
Convert module to Swift

scannillo:
### Motivation

- Swift will enable easier collaboration b/w mobile & web devs.

### Approach

- Leave Demo app & Unit Tests in ObjC to ensure compatibility
- This PR _mostly_ just converts 1:1 as best as possible from Objective-C. Swifty improvements are called out in TODOs and will be addressed in follow-up PRs.

### Changes

- Source code:
    - Convert `POPPopupBridge` --> Swift
    - Convert `POPPopupBridgeDelegate` --> Swift
    - Remove `POPWeakScriptMessageDelegate`, no longer necessary in Swift
   - Add `PopupBridgeUserScript` & `URL+QueryDictionary` helpers

### Next Steps
- Address TODOs left throughout PR
- Docstring audit (make sure all methods have docs & clarify them)
- Remove ObjC `POP` namespace
- Test on device
- Verify Carthage build

 ### Checklist

 - [X] Added a changelog entry

### Authors
@scannillo


jaxdesmarais (commented):
👏 😍 🚀  This is going to make Popup Bridge so much easier to maintain and understand! Thanks for kicking off the conversion and for leaving so many comments and todos for future us.

jaxdesmarais (commented):
Another general question I just thought of, but to we have plans to move this to ASWebAuthenticationSession?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -0,0 +1,194 @@
...
+    /// Handle completion of the popup flow by calling this method from either
+    /// your scene:openURLContexts: scene delegate method or
+    /// your application:openURL:sourceApplication:annotation: app delegate method.
+    @objc(openURL:)
+    public static func open(url: URL) -> Bool {
+        if let returnBlock {
+            return returnBlock(url)
+        } else {
+            return false
+        }
```
sshropshire, line 69:
```suggestion
    returnBlock?(url) ?? false
```
I'm not sure if this is valid, but we could leverage optional syntax more here. Even if this is valid, is the loss of expressiveness worth it? Brevity may decrease readability in this case.

Sources/PopupBridge/PopupBridgeUserScript.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+
+struct PopupBridgeUserScript {
+    let scheme: String
+    let scriptMessageHandlerName: String
+    let host: String
+    
+    var rawJavascript: String {
+        return """
+        ;(function () {
```
sshropshire, line 10: 🤔 : I wonder if that leading `;` does anything? Maybe we could remove.

sshropshire (approved):
🌊 This is awesome!

PopupBridge.podspec
```diff
@@ -19,9 +19,9 @@ Use cases for PopupBridge:
   s.source           = { :git => 'https://github.com/braintree/popup-bridge-ios.git', :tag => s.version.to_s }
 
   s.ios.deployment_target = '9.0'
-
-  s.source_files = 'Sources/PopupBridge/**/*.{h,m}'
-  s.public_header_files = 'Sources/PopupBridge/Public/**/*.h'
+  s.swift_version    = "5.7"
```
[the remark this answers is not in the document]
scannillo: > Should we call this out in the changelog too? I also expect we want to bump this in the Package.swift on line 1.

Done! Good catch.

> Do we also want to bump the Xcode version?

We currently don't specify an Xcode version in the README. I'll save that for a follow-up PR along with several other README cleanups.

Sources/PopupBridge/POPPopupBridge.swift
scannillo: Good call - I have a [TODO](https://github.com/braintree/popup-bridge-ios/blob/swift-conversion/Sources/PopupBridge/POPPopupBridge.swift#L41) to remove this whole separate `returnBlock` concept in a follow-up PR. 

Gonna 🏈 punt on this feedback for that follow up - it'll also probably be obsolete in that follow up.

Sources/PopupBridge/PopupBridgeUserScript.swift
scannillo: Yeh, great question! I'll mess around with this, see if it's necessary. This was just a 1:1 copy of what existed in ObjC.

Will def make a follow-up PR though to simplify that JS snippet if we can.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -0,0 +1,194 @@
...
+        
+        self.dismissSafariViewController()
+        
+        print(urlComponents)
+        
+        var payloadDictionary: [String: Any] = [:]
+        payloadDictionary["path"] = urlComponents.path
+        payloadDictionary["queryItems"] = returnURL.queryDictionary
+        payloadDictionary["hash"] = urlComponents.fragment
```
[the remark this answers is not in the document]
scannillo: Gonna 🏈 punt on this - I left this [TODO](hhttps://github.com/braintree/popup-bridge-ios/pull/45/commits/9500151c75e397753f3eba8aafc9c595812c9eea) for moving to an encodable struct here.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -0,0 +1,194 @@
...
+    private let delegate: POPPopupBridgeDelegate
+    private var safariViewController: SFSafariViewController?
+    
+    private static var scheme: String?
+    private static var returnBlock: ((URL) -> Bool)?
+    
+    // MARK: - Initializer
+    
+    // TODO: - make unfailable
+    public init?(webView: WKWebView, delegate: POPPopupBridgeDelegate) {
```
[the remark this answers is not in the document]
scannillo: 👋 Yeh that's an interesting callout. I have[ this PR](https://github.com/braintree/popup-bridge-ios/pull/46) almost done for addressing the TODOs around requiring the `scheme` and for making the `init` non-failable.

IMO - I feel like I don't see inits that throw a ton in Apple's APIs, so they don't feel super natural to me. But we could do some research.

For this case, I don't think we need an init that throws since PopupBridge _must_ have a url scheme. Let me know if this makes sense!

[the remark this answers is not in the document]
jaxdesmarais: That makes sense - more just a general thought than anything else. You're right they don't use it much in their older frameworks (Foundation as an example) but they do see to use it some of their Swift first frameworks (examples: [file document SwiftUI init](https://developer.apple.com/documentation/swiftui/referencefiledocument/init(configuration:)) and [navigation path](https://developer.apple.com/documentation/swiftui/navigationpath/codablerepresentation/init(from:)) - looks like Airbnb uses it in some places as well like [here](https://github.com/airbnb/lottie-ios/blob/7484b4f4580116d7bf135b5b6097f38f801deaf8/Sources/Private/Model/ShapeItems/Shape.swift#L22) and [here](https://github.com/airbnb/lottie-ios/blob/e1a05adf37fb0557459a2628ef6089571d0bfbc0/Sources/Public/DotLottie/DotLottieFile.swift#L23)).

On the Apple side, it will be interesting to see if this is a pattern they follow or not when moving Foundation to Swift. Not a deal breaker either way, just an observation that we should explore if it makes sense for us at some point.

jaxdesmarais:
> That is a great question! I wasn't planning on it in terms of the Swift re-write, but we definitely can whenever we get around to it. I don't think there's a need to cram it into V2 though since we rarely get time to work on PopupBridge.
> 
> It would save merchants the urlScheme step, too!

True - I wonder if that would also reduce merchant onboarding and confusion. I know there have been some inbounds before around messing up the URL scheme set up. May be a nice quick win to knock out if we have time.

jaxdesmarais (approved):
👏 👏 👏

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # PopupBridge iOS Release Notes
 
+## unreleased
+
+* Convert PopupBridge to Swift
+* Breaking Changes
+  * Remove deprecated `POPPopupBridge.open(url:, sourceApplication:)` & `POPPopupBridge.open(url:, options:)` methods
```
scannillo, line 7: Match Apple doc styling

```suggestion
  * Remove deprecated `POPPopupBridge.open(url:sourceApplication:)` & `POPPopupBridge.open(url:options:)` methods
```

PR #51, opened 2023-04-07
Hide protocol implementations from `POPPopupBridge` public API

scannillo:
### Summary of changes

 - Add `JavaScriptMessageHandler` & `SafariViewControllerEventHandler` classes
     - These classes wrap `WKScriptMessageHandler` and `SFSafariViewControllerDelegate` protocol implementations respectively
     - The purpose is to remove public implementation details for those protocols from the public API for `POPPopupBridge.swift`
    
**Note:**
- This is not something we had to bother with in Objective-C
- Swift requires that if a public class implements a protocol, the protocol implementation methods must also be public. For example:
```
Method 'userContentController(_:didReceive:)' must be declared public because it matches a requirement in public protocol 'WKScriptMessageHandler'
```

 ### Checklist

 - ~ Added a changelog entry~

### Authors
@scannillo


scannillo:
Boo - these errors are because I was using features only avail in Xcode 14.3 ... working on it

scannillo:
I'm closing this PR for a few reasons:
1. The unit tests rely on these protocol implementations to be visible. We don't really have the time right now to be overhauling the unit test setup for popup bridge.
2. This solution is an additional layer of complexity when our goal is just to convert the module to swift for now.

Will follow-up with a simpler PR to just "hide" the methods not meant to be included in the public API. At a later time we can revisit this approach.

PR #56, opened 2023-06-02
Convert UI and Unit Tests to Swift

jaxdesmarais:
### Summary of changes

I started spiking out the changes to use `ASWebAuthenticationSession` in Popup Bridge. Since the tests were in Obj-C it would have required making things public that we don't want public in order to test them in Obj-C. With the tests in Swift it will make the switch to `ASWebAuthenticationSession` easier. All tests are converted 1:1 from Obj-C.

 - Convert `PopupBridge_DemoUITests` to Swift
 - Convert `PopupBridge_UnitTests`
     - Add `MockUserContentController` to replace `OCMock` which is Obj-C only
     - Add `MockScriptMessage` to replace `OCMock` which is Obj-C only
     - Add `MockPopupBridgeDelegate` to replace `OCMock` which is Obj-C only
 - Remove `OCMock` from `Podfile` and `Podfile.lock` - this pod is no longer in use

 ### Checklist

- @jaxdesmarais


sshropshire (approved):

UnitTests/MockScriptMessage.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+import WebKit
+
+class MockScriptMessage: WKScriptMessage {
+
+    var _body: Any?
+    var _name: String?
```
scannillo, line 7: Can we make these private?

UnitTests/PopupBridge_UnitTests.swift
```diff
@@ -0,0 +1,280 @@
...
+        let _ = POPPopupBridge(webView: webView, urlScheme: returnURL, delegate: delegate)
+
+        XCTAssertEqual(webView.configuration.userContentController.userScripts.count, 1)
+
+        let userScript = webView.configuration.userContentController.userScripts[0]
+
+        XCTAssertEqual(userScript.injectionTime, WKUserScriptInjectionTime.atDocumentStart)
+        XCTAssertTrue(userScript.isForMainFrameOnly)
+    }
+
```
scannillo, line 28:
```suggestion
```

UnitTests/MockScriptMessage.swift
jaxdesmarais: Sure can! done here: https://github.com/braintree/popup-bridge-ios/pull/56/commits/a665b15341762a1dd78d9be19e242e29a42bbcb5

PR #57, opened 2023-06-02
Replace `SFSafariViewController` with `ASWebAuthenticationSession`

jaxdesmarais:
### Summary of changes

 - Remove the need to register a URL type to use Popup Bridge
 - Remove `POPPopupBridgeDelegate` methods that apply only to `SFSafariViewController`
 - Add `WebAuthenticationSession`
 - Replace `SFSafariViewController` with `ASWebAuthenticationSession`
 - Update README with new integration steps
 - Remove code paths related to `SFSafariViewController`
 - Update UnitTests

 ### Checklist

 - [x] Added a changelog entry

- @jaxdesmarais


README.md
```diff
@@ -26,9 +26,6 @@ To integrate using [CocoaPods](https://cocoapods.org), add the following line to
 ```ruby
 pod 'PopupBridge'
 ```
-### Carthage
```
jaxdesmarais, line 29: None of our release or build steps build or release frameworks for Carthage. Seems like we don't actually support Carthage unless I am missing something?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -1,63 +1,72 @@
...
-    
+    private let hostName = "popupbridgev1"    
     private let webView: WKWebView
-    private let urlScheme: String
     private let delegate: POPPopupBridgeDelegate
-    private var safariViewController: SFSafariViewController?
+    private var webAuthenticationSession: WebAuthenticationSession
     
-    private static var returnBlock: ((URL) -> Bool)?
+    private var returnBlock: ((URL) -> Void)? = nil
```
jaxdesmarais, line 19: It doesn't seem like there is a specific reason we need to return a bool but I may have missed a use case for this. Everything is working as expected in the demo app. The only exception is the Ideal flow. Once this is merged we should update the [example here to pass in the updated url scheme](https://github.com/braintree/popup-bridge-example/blob/master/local-payment-methods.html#L78).

README.md
```diff
@@ -44,38 +41,6 @@ To run the sample app, clone the repo, open `PopupBridge.xcworkspace` and run th
 Quick Start
 -----------
 
-1. Register a URL type for your app:
```
jaxdesmarais, line 47: goodbye gif

README.md
scannillo: You're right we don't verify anywhere that `carthage update` succeeds, but it very well could be used with the SDK still.

Sources/PopupBridge/POPPopupBridgeDelegate.swift
```diff
@@ -2,16 +2,6 @@ import UIKit
...
-    @objc func popupBridge(_ bridge: POPPopupBridge, requestsPresentationOfViewController viewController: UIViewController)
-
-    /// You must dismiss the view controller in this method.
-    @objc func popupBridge(_ bridge: POPPopupBridge, requestsDismissalOfViewController viewController: UIViewController)
-    
-    /// Optional: Receive the URL that will be opened
-    @objc optional func popupBridge(_ bridge: POPPopupBridge, willOpenURL url: URL)
 
     /// Optional: Receive messages from the web view
     @objc optional func popupBridge(_ bridge: POPPopupBridge, receivedMessage messageName: String, data: String?)
```
scannillo, line 7: I don't think we need `POPPopupBridgeDelegate` at all anymore. The optional `receivedMessage()` function was really just a nice-to-have by product of our SDKs communication with `WKScriptMessageHandler` that we decided to surface to the merchant in case they wanted it. Since it isn't necessary for a PopupBridge integration, I think we should remove it.

UnitTests/MockWebAuthenticationSession.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+import AuthenticationServices
+@testable import PopupBridge
+
+class MockWebAuthenticationSession: WebAuthenticationSession {
```
scannillo, line 5: This is also another case for breaking out our ASWeb wrapper from BTCore into it's own module - PopupBridge could consume it too. 🏈

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -160,22 +145,47 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler, SFSafariViewContr
...
-                let viewController = SFSafariViewController(url: url)
-                safariViewController = viewController
-                safariViewController?.delegate = self
-                
-                self.delegate.popupBridge(self, requestsPresentationOfViewController: viewController)
-                return
+            if let urlString = script.url, let url = URL(string: urlString) {
+                webAuthenticationSession.start(url: url, context: self) { url, error in
+                    if let error = error as? NSError {
+                        if error.code == ASWebAuthenticationSessionError.canceledLogin.rawValue {
```
scannillo, line 151: Can we move the cancel logic into it's own callback param like we did on BT?

README.md
jaxdesmarais: Wouldn't we still need to distribute the framework for Carthage? For example, when we run `zip -r Braintree.xcframework.zip Carthage` in our BT release, once that's done we upload that zip to the release it shows up here: https://github.com/braintree/braintree_ios/releases/tag/5.22.0 as a downloadable zip. With popup bridge, we don't do these steps and there is no zip of the framework here: https://github.com/braintree/popup-bridge-ios/releases/tag/1.2.0

Again, maybe I've totally misunderstood Carthage magic this while time but my understanding was that we'd need that zip of the framework to exist in order for a merchant to use the SDK with Carthage.

Sources/PopupBridge/POPPopupBridgeDelegate.swift
jaxdesmarais: Done: https://github.com/braintree/popup-bridge-ios/pull/57/commits/93c4c3662227d1b5519859b0555507cd59f81505 always down to 🔥

Sources/PopupBridge/POPPopupBridge.swift
jaxdesmarais: 👍 https://github.com/braintree/popup-bridge-ios/pull/57/commits/abef023b316916f5da4bdbd38626214e9abb3af6

README.md
scannillo: Ah good question - so Carthage doesn't require those `zip` files be attached to the releases. Those are a nice-to-have feature that speeds up merchants time pulling down the SDK via Carthage. We used to not ship them with our releases for awhile, but a merchant requested we add it awhile ago (https://github.com/braintree/braintree_ios/issues/543). 

(Also - If a Carthage user wants to bypass using our pre-built binaries, they can pass the flag `carthage update --no-use-binaries`)

scannillo: So since this SDK has no dependencies, it doesn't need an SDK-level Cartfile. A merchant could still be including this SDK in their app via Carthage by specifying our URL in their Cartfile and building from source (like: `github "braintree/popup-bridge-ios" ~> 5.5`).

jaxdesmarais: 🤯 learn something new about package managers every day! Thanks for the explanation as always, this whole time I assumed that zipping the binaries was required to use Carthage. Doesn't seem like we've gotten any inbounds about it on this SDK, assuming because it's so small, but we could always add it here in the future to have consistencies across all the SDKs.

jaxdesmarais: Reverted this here: https://github.com/braintree/popup-bridge-ios/pull/57/commits/f7c255efbfbc3de1de2543e00182b1894204610d

CHANGELOG.md
```diff
@@ -7,7 +7,10 @@
   * Require iOS 14+, Xcode 14.3+, and Swift 5.8+
   * Remove deprecated `POPPopupBridge.open(url:sourceApplication:)` & `POPPopupBridge.open(url:options:)` methods
   * Remove `POPPopupBridge.set(returnURLScheme:)`
-  * Add required `urlScheme` parameter to `POPPopupBridge.init()`
+  * Replace `SFSafariViewController` with `ASWebAuthenticationSession`
+  * No longer need to register a URL type to use Popup Bridge
+  * Removed `POPPopupBridgeDelegate`
+  * The initializer for `POPPopupBridge` now only requires a `WKWebView` `POPPopupBridge(webView:)`
```
scannillo, line 13: Nitpick

```suggestion
  * The initializer for `POPPopupBridge` now only requires a `WKWebView`: `POPPopupBridge(webView:)`
```

README.md
```diff
@@ -191,20 +145,9 @@ However, WKWebViews do not display an address bar or an HTTPS lock icon. If the
 
 - PopupBridge attaches to a WKWebView by [injecting a user script](https://developer.apple.com/reference/webkit/wkusercontentcontroller/1537448-adduserscript) to the page
   - This exposes a JavaScript interface (via `window.popupBridge`) for the web page to interact with the iOS code
-- The web page detects whether the page has access to `window.popupBridge`; if so, it uses `popupBridge.open` to open the popup URL
-  - `popupBridge.open` creates a SFSafariViewController to open the popup URL and has its delegate present the view controller
+- The web page detects whether the page has access to `window.popupBridge`; if so, it creates a ASWebAuthenticationSession to open the popup URL
   - The web page can also use `popupBridge.onComplete` as a callback
-- The popup web page uses a deep link URL to dismiss the popup
```
scannillo, line 197: Did we check to see if a cancellation that comes from within the pop-up (such as the tiny cancelation text in the PP flow) properly cancels?

scannillo: We should check to see if any changes need to be made to the JS portion of the PopopBridge integration

Sources/PopupBridge/POPPopupBridge.swift
scannillo: > Once this is merged we should update the [example here to pass in the updated url scheme](https://github.com/braintree/popup-bridge-example/blob/master/local-payment-methods.html#L78).

The JS integration shouldn't care about a URL scheme anymore, though for the new ASWeb version of PopupBridge?

scannillo: > It doesn't seem like there is a specific reason we need to return a bool but I may have missed a use case for this.

The bool is returned to the merchant after they call[ `open(url:)` ](https://github.com/braintree/popup-bridge-ios/blob/844157c1de77834987c9aca32c4d3c79fc821cb9/Sources/PopupBridge/POPPopupBridge.swift#L56-L62), which looks to be removed in this PR. We should call out this method was removed in the CHANGELOG.

README.md
jaxdesmarais: > Did we check to see if a cancellation that comes from within the pop-up (such as the tiny cancelation text in the PP flow) properly cancels?

Yep, tested that in cancels/dismisses the window and hits the same block that it did in 1.x branch.

Sources/PopupBridge/POPPopupBridge.swift
jaxdesmarais: > The JS integration shouldn't care about a URL scheme anymore, though for the new ASWeb version of PopupBridge?

Yep, but since we have a fallback URL hardcoded it's always attempting to fall back to that URL scheme, which doesn't exist. We _should_ be able to remove it completely but will also need to make sure Android popup bridge doesn't rely on that being hardcoded.

jaxdesmarais: > The bool is returned to the merchant after they call[ open(url:) ](https://github.com/braintree/popup-bridge-ios/blob/844157c1de77834987c9aca32c4d3c79fc821cb9/Sources/PopupBridge/POPPopupBridge.swift#L56-L62), which looks to be removed in this PR. We should call out this method was removed in the CHANGELOG.

Ah, do we need an additional line? I had assumed this would be covered by the [following line already in the CHANGELOG](https://github.com/braintree/popup-bridge-ios/blob/master/CHANGELOG.md#unreleased): `Remove deprecated POPPopupBridge.open(url:sourceApplication:) & POPPopupBridge.open(url:options:) methods`. It didn't look like we had called out the rename to `open(url:)` anywhere but happy to add an additional line if it's not clear!

jaxdesmarais: Ah 🤦  - I thought it was a rename and not 3 separate methods. Updated here: https://github.com/braintree/popup-bridge-ios/pull/57/commits/38eede3be39a3e165e2e24368a545fa00a612bfa

README.md
```diff
@@ -191,20 +145,9 @@ However, WKWebViews do not display an address bar or an HTTPS lock icon. If the
...
-  - The app delegate handles the deep link URL and forwards it to PopupBridge
-  - One way to avoid hard-coding the deep link is by adding it as a query parameter to the popup URL:
-
-    ```javascript
-      popupBridge.open(url + '?popupBridgeReturnUrlPrefix=' + popupBridge.getReturnUrlPrefix());
-    ```
-
-    - Optionally, you can add path components and query parameters to the deep link URL to return data to the parent page, which are provided in the payload of `popupBridge.onComplete` 
-- If the user taps the **Done** button on the SFSafariViewController, `popupBridge.onComplete` gets called with the error and payload as `null` and the delegate dismisses the view controller
+- If the user taps the **Done** button on the ASWebAuthenticationSession, `popupBridge.onComplete` gets called with the error and payload as `null`
```
scannillo, line 150:
```suggestion
- If the user taps the **Cancel** button on the ASWebAuthenticationSession, `popupBridge.onComplete` gets called with the error and payload as `null`
```

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -1,63 +1,61 @@
...
-            return false
+
+    /// Exposed for testing
+    init(webView: WKWebView, webAuthenticationSession: WebAuthenticationSession) {
+        self.webView = webView
+        self.webAuthenticationSession = webAuthenticationSession
+
+        super.init()
+
+        configureWebView(scheme: PopupBridgeConstants.callbackURLScheme)
```
scannillo, line 50: The `configureWebView` func signature doesn't need to take a `scheme` anymore, it can just access `PopupBridgeConstants.callbackURLScheme` directly

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -1,63 +1,61 @@
...
-    /// your application:openURL:sourceApplication:annotation: app delegate method.
-    @objc(openURL:)
-    public static func open(url: URL) -> Bool {
-        if let returnBlock {
-            return returnBlock(url)
-        } else {
-            return false
+
+    /// Exposed for testing
+    init(webView: WKWebView, webAuthenticationSession: WebAuthenticationSession) {
```
scannillo, line 44: Ideally we avoid having duplicate logic just for dependency injection for tests. This opens the door for errors since we have to maintain 2 inits to be the exact same and the real one isn't actually being tested.

We could do something like this:
```
    // MARK: - Private Properties
    
    ...
    
    // MARK: - Initializers
        
    /// Merchant facing init
    public init(webView: WKWebView) {
        self.webView = webView
        
        super.init()

        configureWebView(scheme: PopupBridgeConstants.callbackURLScheme)
                
        returnBlock = { url in
            guard let script = self.constructJavaScriptCompletionResult(returnURL: url) else {
                return
            }

            self.injectWebView(webView: webView, withJavaScript: script)
            return
        }
    }

    /// Exposed for testing
    convenience init(webView: WKWebView, webAuthenticationSession: WebAuthenticationSession) {
        self.init(webView: webView)
    }
```

scannillo: Or, not have a 2nd init at all and just expose the `webAuthSession` property. 

```
    // MARK: - Internal Properties
    
    /// Exposed for testing
    var webAuthenticationSession: WebAuthenticationSession = WebAuthenticationSession()

    // MARK: - Initializers
        
    /// Merchant facing init
    public init(webView: WKWebView) {
        self.webView = webView
        
        super.init()

        configureWebView(scheme: PopupBridgeConstants.callbackURLScheme)
                
        returnBlock = { url in
            guard let script = self.constructJavaScriptCompletionResult(returnURL: url) else {
                return
            }

            self.injectWebView(webView: webView, withJavaScript: script)
            return
        }
    }
```

I think the 2nd init is a _little_ more clear and less error prone in the test suite, but I could go either way. What do you think?

Sources/PopupBridge/POPPopupBridge.swift
jaxdesmarais: Updated: https://github.com/braintree/popup-bridge-ios/pull/57/commits/133f5611b1f883393b2d38a14ac4978b2cfabd09

Sources/PopupBridge/POPPopupBridge.swift
jaxdesmarais: Yeah I am down for this approach to reduce the duplication!

jaxdesmarais: ✅ : https://github.com/braintree/popup-bridge-ios/pull/57/commits/1391a3bd090a2b0902dc5d78c7f123f13fd5478d

scannillo (approved):
👏 Awesome! Only thing is that before we release this we need a corresponding MIGRATION guide update.

PR #60, opened 2023-08-25
Fix various memory leaks caused by capturing self

jhong70:
### Summary of changes

 - There are various points in POPPopupBridge.swift where self is captured and retained. This causes POPPopupBridge and web views to leak. This change addresses those issues by passing in self as weak reference. Additionally we prevent leaks from using WKScriptMessageHandler by creating a proxy object that weekly retains the handler.

 ### Checklist

 - [x] Added a changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## unreleased
 
 * Convert PopupBridge to Swift
+* Fix memory leaks in POPPopupBridge
 * Breaking Changes
   * Require iOS 14+, Xcode 14.3+, and Swift 5.8+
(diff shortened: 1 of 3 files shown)


scannillo:
Hi @jhong70, thank you for opening this PR! Can you provide some more specifics on between which two objects you're seeing a retain cycle as well as steps for how we can replicate this leak?

via-guy:
@scannillo I recently fixed a memory leak in my project for `webView.configuration.userContentController.add`, which retains the handler. It was very difficult to debug.

jaxdesmarais:
Closing due to inactivity - can reopen with steps to reproduce so we can verify the fix

PR #75, opened 2025-03-04
[V3] Add Networking to Send Analytics

richherrera:
### Summary of changes

 - 

Screenshot 2025-03-04 at 12 10 40 p m

 ### Checklist

 - [ ] Added a changelog entry

### Authors

- @richherrera


Sources/PopupBridge/NetworkClient.swift
```diff
@@ -0,0 +1,40 @@
...
+
+final class NetworkClient: Networkable {
+    
+    // MARK: - Private Properties
+    
+    private let session: Sessionable
+    
+    // MARK: - Initializer
+    
+    init(session: Sessionable = URLSession.shared) {
```
jaxdesmarais, line 15: Thoughts on just hardcoding `URLSession.shared` here vs using the `Sessionable` protoocol? Not sure if it's needed for testing.

Sources/PopupBridge/NetworkClient.swift
```diff
@@ -0,0 +1,40 @@
...
+        self.session = session
+    }
+    
+    // MARK: - Internal Methods
+    
+    func post<T: Encodable>(url: URL, body: T) async throws {
```
jaxdesmarais, line 21: Not sure if it makes it more testable to strongly type here vs using a generic. We have no other use for our networking in this SDK.

Sources/PopupBridge/NetworkClient.swift
```diff
@@ -0,0 +1,40 @@
...
+        var request = URLRequest(url: url)
+        request.httpMethod = "POST"
+        request.allHTTPHeaderFields = ["Content-Type": "application/json"]
+        
+        do {
+            let encodedBody = try JSONEncoder().encode(body)
+            request.httpBody = encodedBody
+        } catch let encodingError {
+            throw NetworkError.encodingError(encodingError)
+        }
```
jaxdesmarais, line 31: I think we can use the shorthand here that swift gives us for free:
```suggestion
        } catch {
            throw NetworkError.encodingError(error)
        }
```

Sources/PopupBridge/AnalyticsService.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+protocol AnalyticsServiceable {
+    func sendAnalyticsEvent(_ eventName: String, sessionID: String)
```
warmkesselj, line 4: Curious why we need this protocol?

Sources/PopupBridge/NetworkClient.swift
richherrera: Updated: [9a5e980](https://github.com/braintree/popup-bridge-ios/pull/75/commits/9a5e980c1a902a7c03f32083c25e6f3428cc4b30) since we have a strongly typed, this catch is unnecessary

Sources/PopupBridge/NetworkClient.swift
richherrera: sure, updated: [9a5e980](https://github.com/braintree/popup-bridge-ios/pull/75/commits/9a5e980c1a902a7c03f32083c25e6f3428cc4b30)

Sources/PopupBridge/AnalyticsService.swift
richherrera: Thanks for asking, this protocol acts as the contract. With this protocol, I can utilize dependency injection ([here](https://github.com/braintree/popup-bridge-ios/blob/9a5e980c1a902a7c03f32083c25e6f3428cc4b30/Sources/PopupBridge/POPPopupBridge.swift#L11)), which allows me to create a mock of the AnalyticsService class, enabling me to perform unit tests on the PopupBridge class

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -8,13 +8,16 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
     /// Exposed for testing
     var returnedWithURL: Bool = false
     
+    static var analyticsService: AnalyticsServiceable = AnalyticsService()
```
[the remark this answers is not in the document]
richherrera: Excellent observation, let me check. The reason i created this static property is because we need to send the `started` event when this class is instantiated. I couldn't find a way to use dependency injection, and this approach allows me to conduct the necessary tests

Sources/PopupBridge/AnalyticsService.swift
```diff
@@ -0,0 +1,42 @@
...
+    
+    func sendAnalyticsEvent(_ eventName: String, sessionID: String) {
+        Task(priority: .background) {
+            await performEventRequest(eventName, sessionID: sessionID)
+        }
+    }
+    
+    func performEventRequest(_ eventName: String, sessionID: String) async {
+        let body = createAnalyticsEvent(eventName: eventName, sessionID: sessionID)
+        try? await networkClient.post(url: url, body: body)
```
warmkesselj, line 31: Thoughts on do catch here to surface the errors with a log?

Sources/PopupBridge/FPTIBatchData.swift
```diff
@@ -2,77 +2,110 @@ import UIKit
...
+        
+        let merchantAppVersion: String = Bundle.main.infoDictionary?[kCFBundleVersionKey as String] as? String ?? "N/A"
+        
+        let packageManager: String = {
+            #if COCOAPODS
+                "CocoaPods"
+            #elseif SWIFT_PACKAGE
+                "Swift Package Manager"
+            #else
+                "Carthage or Other"
```
warmkesselj, line 87: I believe project is setup with Cocoapods and I'm observing `Carthage or Other` being returned.

richherrera: This is part of the parameters sent in the events, helping to identify which `dependency manager` our merchant uses. Our demo app doesn't use any since it's within the same `workspace`, and this implementation is based on the implementation in [BT](https://github.com/braintree/braintree_ios/blob/7e8b41c17bc23be808c27805ef7fbd1ca1be1b4d/Sources/BraintreeCore/Analytics/FPTIBatchData.swift#L168) and [PPCP](https://github.com/paypal/paypal-ios/blob/d5f1f5c1acc8733c7299893a8cb57e899ed81c0b/Sources/CorePayments/AnalyticsEventData.swift#L61).

Sources/PopupBridge/AnalyticsService.swift
warmkesselj: Are we expecting other Service types besides AnalyticsServiceable? And is this thinking leading to the potential creation of  a base class or base models that have a `static var analyticsService: AnalyticsServiceable = AnalyticsService()` property?

warmkesselj: What if the class itself conformed to `AnalyticsServiceable`?

Sources/PopupBridge/AnalyticsService.swift
richherrera: Sure Updated: [a658633](https://github.com/braintree/popup-bridge-ios/pull/75/commits/a658633a92b817d8eb13ab60665a5f832d043e2e)

same as we have on [PPCP](https://github.com/paypal/paypal-ios/blob/d5f1f5c1acc8733c7299893a8cb57e899ed81c0b/Sources/CorePayments/Networking/AnalyticsService.swift#L74)

Sources/PopupBridge/AnalyticsService.swift
richherrera: No, we are not expecting for other services, which is why the NetworkClient class only accepts the FPTIBatchData model

If we have another class that needs to send events, we would simply use dependency injection, depending on the analysis at the time. This property wouldn't need to be static; it would essentially be the same as what I did [here](https://github.com/braintree/popup-bridge-ios/blob/a658633a92b817d8eb13ab60665a5f832d043e2e/Sources/PopupBridge/AnalyticsService.swift#L17). 

If the PopupBridge class conforms to this protocol, we would have to move all the logic that is in AnalyticsService, which means mixing responsibilities, and the network layer (NetworkClient) would be the static property. Happy to pair and discuss in more detail.

warmkesselj: Yeah, pairing sounds good. I'll DM.

Sources/PopupBridge/FPTIBatchData.swift
```diff
@@ -2,77 +2,110 @@ import UIKit
 
 struct FPTIBatchData: Codable {
     
-    let appID: String = Bundle.main.infoDictionary?[kCFBundleIdentifierKey as String] as? String ?? "N/A"
```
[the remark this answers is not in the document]
jaxdesmarais: That's true. Up to you, we send so few events in this SDK that it may not be worth batching. In BT we did used to send single events to the batch endpoint (see https://github.com/braintree/braintree_ios/pull/1295/files), so we could always do that and add batching to this SDK if we find it useful in the future. What do you think? I don't feel strongly one way or the other so will defer to you.

Sources/PopupBridge/POPPopupBridge.swift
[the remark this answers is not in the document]
jaxdesmarais: 🤔 yeah we should definitely address it not deinitializing or provide merchants more explicit instructions on destroying this class. If it's happening on main as well I don't want to make it a blocker, but can you make a ticket for us to dig into this a bit more post release?

Should we consider moving the started event out of the init and into the first thing that's called post init? Would that allow us to make this non static? Sounds like we'd still want to address the deinit of this class at a later time.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -76,10 +84,12 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
 
         if let payloadData = try? JSONEncoder().encode(payload),
            let payload = String(data: payloadData, encoding: .utf8) {
+            Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.succeeded, sessionID: sessionID)
```
jaxdesmarais, line 87: Take it or leave it, we could make `sendAnalyticsEvent` take in a `PopupBridgeAnalytics` as the first argument, then just use:
```suggestion
            Self.analyticsService.sendAnalyticsEvent(.succeeded, sessionID: sessionID)
```

Sources/PopupBridge/POPPopupBridge.swift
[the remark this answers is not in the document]
richherrera: Ticket DTMOBILES-1487

Maybe i can think of another approach. We have two flows: PayPal (via PayPal -> Pay with Credit Card) and Venmo, where there isn't another method called after the init that we could use to send the 'start' event.

Sources/PopupBridge/POPPopupBridge.swift
jaxdesmarais: Ah you're right, forgot these were caseless enums.

jaxdesmarais (approved):
One small docstrings cleanup, but assuming we verified we are seeing these events in FPTI, this looks great to me! 🚀

UnitTests/MockNetworkClient.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+@testable import PopupBridge
```
warmkesselj, line 2: Is this import being used or is it necessary here?

Sources/PopupBridge/NetworkError.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+enum NetworkError: Error {
+    case invalidResponse
```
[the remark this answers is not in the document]
richherrera: I think this way it remains as basic as possible, which is what's expected, since  im using this in the NetworkClient, the scope of httpResponse is limited within the guard block, so i cannot directly use it in the throw statement outside of that block. If in the future we need to add more endpoints, I think that would add more value.

PopupBridge.xcodeproj/project.pbxproj
```diff
@@ -7,10 +7,20 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		454C7EC32D6E351F00CA3191 /* NetworkClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 454C7EC22D6E351900CA3191 /* NetworkClient.swift */; };
+		45AE41D22D7649F800388548 /* MockAnalyticsService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45AE41D12D7649EE00388548 /* MockAnalyticsService.swift */; };
 		45CD0C2C2D64F08F0072C5A4 /* FPTIBatchData.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45CD0C2B2D64F0810072C5A4 /* FPTIBatchData.swift */; };
 		45CD0C2E2D664D140072C5A4 /* Date+MilisecondTimestamp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45CD0C2D2D664CF50072C5A4 /* Date+MilisecondTimestamp.swift */; };
 		45CD0C302D6788A10072C5A4 /* Bundle+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45CD0C2F2D67888F0072C5A4 /* Bundle+Extension.swift */; };
 		45CD0C322D6793FB0072C5A4 /* PopupBridgeAnalytics.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45CD0C312D6793F90072C5A4 /* PopupBridgeAnalytics.swift */; };
+		45FBAF272D6F9CCF000D550B /* AnalyticsService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45FBAF262D6F9CC6000D550B /* AnalyticsService.swift */; };
```
scannillo, line 16: Can we add all of the new files that were specifically added for analytics to an `Analytics` folder under `Sources/PopupBridge`? nitpick, but would help navigation of the PopupBridge source files

Sources/PopupBridge/POPPopupBridge.swift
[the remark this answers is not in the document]
scannillo: Why does it need to be static? Can it just be an instance property?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -8,13 +8,16 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
     var returnedWithURL: Bool = false
     
+    static var analyticsService: AnalyticsServiceable = AnalyticsService()
+    
     // MARK: - Private Properties
     
     private let messageHandlerName = "POPPopupBridge"
-    private let hostName = "popupbridgev1"    
+    private let hostName = "popupbridgev1"
+    private let sessionID = UUID().uuidString.replacingOccurrences(of: "-", with: "")
```
scannillo, line 17: Can this sessionID creation live in AnalyticsService instead? Since it isn't required by any of the login in POPPopupBridge.swift

Sources/TestPlan.xctestplan
```diff
@@ -0,0 +1,24 @@
+{
```
scannillo, line 1: Can this file be deleted?

Sources/PopupBridge/AnalyticsService.swift
```diff
@@ -0,0 +1,46 @@
...
+protocol AnalyticsServiceable {
+    func sendAnalyticsEvent(_ eventName: String, sessionID: String)
+}
+
+final class AnalyticsService: AnalyticsServiceable {
+    
+    // MARK: - Private Properties
+    
+    /// The FPTI URL to post all analytic events.
+    private let url = URL(string: "https://api.paypal.com/v1/tracking/batch/events")!
```
scannillo, line 12: Since we aren't actually batching, FPTI has a separate endpoint for single event upload `/v1/tracking/events/`. The format is slightly different from `/v1/tracking/batch/events/`. I'm not sure if there are performance benefits to using the single event API vs the multiple event API, but we could confirm in #help-fpti.

To see the structure of each, you can look at the API details in PPAAS (I also have it in a Postman collection)

Sources/PopupBridge/POPPopupBridge.swift
richherrera: At the moment, no, because the sessionID would always be the same, due to AnalyticsService being a static var.

If you have any suggestions for adding AnalyticsService that isn't a static var, I'd be happy to work on it. The reason it's a static variable is that in tests, I need to inject the mock of AnalyticsService. I can't inject it in the init, and the "start" event is triggered in the init :(

Sources/PopupBridge/POPPopupBridge.swift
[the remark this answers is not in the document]
richherrera: The "start" event is triggered in the public init, and there's no way to inject the mock AnalyticsService for testing. Any suggestions?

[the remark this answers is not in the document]
richherrera: Please take a look at the public init, 

https://github.com/braintree/popup-bridge-ios/blob/70bb98e396ae7f09bb23b34f16f1173f473985e0/Sources/PopupBridge/POPPopupBridge.swift#L36

``` swift
    public init(webView: WKWebView, prefersEphemeralWebBrowserSession: Bool = true) {
        self.webView = webView
        
        super.init()
        
        Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.started, sessionID: sessionID)
        
        configureWebView()
        webAuthenticationSession.prefersEphemeralWebBrowserSession = prefersEphemeralWebBrowserSession
                
        returnBlock = { url in
            guard let script = self.constructJavaScriptCompletionResult(returnURL: url) else {
                return
            }

            self.injectWebView(webView: webView, withJavaScript: script)
            return
        }
    }
```

PopupBridge.xcodeproj/project.pbxproj
richherrera: Updated: [c8fd4cf](https://github.com/braintree/popup-bridge-ios/pull/75/commits/c8fd4cfeae501954350b8081c66e2e04f5f5e95e)

Sources/TestPlan.xctestplan
richherrera: yeap, deleted: [5911546](https://github.com/braintree/popup-bridge-ios/pull/75/commits/591154671a7e84f429bd4d099c715fc2bb03637c)

Sources/PopupBridge/AnalyticsService.swift
richherrera: Yes, it's the endpoint used in PPCP, however, we would need a [custom encode method](https://github.com/paypal/paypal-ios/blob/d5f1f5c1acc8733c7299893a8cb57e899ed81c0b/Sources/CorePayments/AnalyticsEventData.swift#L120). I did a POC with those changes and was thinking that it would be better to have the same implementation we have in BT, which is why I opted for this implementation. While working on BT v7, we chose not to use a custom encode method. Do you think we could use it for this repo?

jaxdesmarais (approved):

PR #79, opened 2025-03-20
[V3] Fix Retain Cycles

richherrera:
### Summary of changes

 - Remove all script handlers in deinit to avoid retain cycles in WebAuthenticationSession.
 - Instead of POPPopupBridge handling WKWebView messages directly, WebViewScriptHandler now acts as an intermediary.
 - WebViewScriptHandler receives messages from WKWebView and forwards them to proxy.
 - The proxy property in WebViewScriptHandler is weak, helping prevent potential retain cycles.
 - Add Unit Test

🐛 
<video src=https://github.com/user-attachments/assets/6cab6599-cfcd-4ea8-84f3-305b1d4f4b1a width="300">  

🧼 
<video src=https://github.com/user-attachments/assets/18225380-c4ac-4b49-be91-553f2ba6d473  width="300"> 

 ### Checklist

### Authors

- @richherrera


jaxdesmarais (approved):
Are we able to add a unit test for this? Not sure if it's easily testable now that it's fixed but not a blocker.

richherrera:
> Are we able to add a unit test for this? Not sure if it's easily testable now that it's fixed but not a blocker.

Sure, I added a new method to check for memory leaks. What I did was add the test and the extension in v3, then did the same in this branch. It seems the tests were successful.

PR #82, opened 2025-05-29
Move Started Event to Match Android

jaxdesmarais:
### Summary of changes

 - Move started event to match Android implementation ([source code](https://github.com/braintree/popup-bridge-android/blob/28c04a0b483566a450e6152de6acd69c767e3853/PopupBridge/src/main/java/com/braintreepayments/api/PopupBridgeClient.kt#L143))

 ### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/PopupBridge/POPPopupBridge.swift
+++ b/Sources/PopupBridge/POPPopupBridge.swift
@@ -100,6 +100,7 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
     /// Injects custom JavaScript into the merchant's webpage.
     /// - Parameter scheme: the url scheme provided by the merchant
     private func configureWebView() {
+        Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.started, sessionID: sessionID)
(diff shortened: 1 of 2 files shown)


Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -144,6 +142,7 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
             }
             
             if let urlString = script.url, let url = URL(string: urlString) {
+                Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.started, sessionID: sessionID)
```
richherrera, line 145: Are we sure this event gets sent when the user chooses to pay with Venmo?

richherrera: I think that makes sense, if it's to stay aligned with Android, I'm totally on board with it.

richherrera (approved):

warmkesselj (approved):

PR #90, opened 2026-03-20
[Feature] PopupBridge++

jtanya17:
### Summary of changes

- Add PayPal app switch support to PopupBridge
- Introduces an optional `enablePopupBridgeAppSwitch` flag to `POPPopupBridge` that, when enabled, allows the SDK to launch the native PayPal app for checkout instead of opening a browser session. 
- When the app launch succeeds, a `NotificationCenter` observer handles the deep link return URL and injects the result back into the WebView. If the app launch fails, the flow falls back to the existing `WebAuthenticationSession` behavior. 
- Exposes `isPayPalInstalled`, `launchApp()`, and `getDeepLinkReturnUrlPrefix()` to JavaScript, and fires analytics events for app detection and launch outcomes.

 ### Checklist

 - [ ] Added a changelog entry

- @jtanya17

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,12 +1,25 @@
 ### Summary of changes
+- 
 
- - 
+### AI Usage
 
- ### Checklist
+**Which AI Agent Was Used?**
+- [ ] Copilot 
+- [ ] Claude 
+- [ ] Other (Type Name Here)
 
- - [ ] Added a changelog entry
+**How was AI used?**
+<!-- Describe how AI was used (e.g., code generation, refactoring, debugging, writing tests, etc.) -->
+
+**Estimated AI Code Contribution**
+- [ ] less than 30% 
+- [ ] 30 - 60% 
+- [ ] 60 - 100% 
+
+### Checklist
+- [ ] Added a changelog entry
+- [ ] Tested and confirmed payment flows affected by this change are functioning as expected
 
 ### Authors
 > List GitHub usernames for everyone who contributed to this pull request.
-
-- 
+-
--- a/IntegrationTests/AnalyticsService_IntegrationTests.swift
+++ b/IntegrationTests/AnalyticsService_IntegrationTests.swift
@@ -1,0 +1,74 @@
+import XCTest
+@testable import PopupBridge
+
+final class AnalyticsService_IntegrationTests: XCTestCase {
+
+    var sut: AnalyticsService!
+
+    let eventName = "test-integration-event"
+    let sessionID = "test-integration-session-id"
+    let analyticsURL = URL(string: "https://api.paypal.com/v1/tracking/batch/events")!
+
+    override func setUp() {
+        super.setUp()
+        StubURLProtocol.reset()
+        let config = URLSessionConfiguration.ephemeral
+        config.protocolClasses = [StubURLProtocol.self]
+        sut = AnalyticsService(session: URLSession(configuration: config))
+    }
+
+    override func tearDown() {
+        sut = nil
+        StubURLProtocol.reset()
+        super.tearDown()
+    }
+
+    // MARK: - performEventRequest
+
+    func test_performEventRequest_success_postsToCorrectEndpointWithCorrectPayload() async {
+        StubURLProtocol.stubbedStatusCode = 200
+
+        await sut.performEventRequest(eventName, sessionID: sessionID)
+
+        XCTAssertEqual(StubURLProtocol.lastRequest?.url, analyticsURL)
+        XCTAssertEqual(StubURLProtocol.lastRequest?.httpMethod, "POST")
+        XCTAssertEqual(StubURLProtocol.lastRequest?.allHTTPHeaderFields?["Content-Type"], "application/json")
+
+        guard let body = StubURLProtocol.lastRequest?.httpBody,
+              let decoded = try? JSONDecoder().decode(FPTIBatchData.self, from: body) else {
+            XCTFail("Request body could not be decoded as FPTIBatchData")
+            return
+        }
+        XCTAssertEqual(decoded.events.first?.fptiEvents.first?.eventName, eventName)
+        XCTAssertEqual(decoded.events.first?.metadata.sessionID, sessionID)
+    }
+
+    func test_performEventRequest_nonSuccessResponse_doesNotPropagateError() async {
+        StubURLProtocol.stubbedStatusCode = 400
+
+        await sut.performEventRequest(eventName, sessionID: sessionID)
+
+        XCTAssertNotNil(StubURLProtocol.lastRequest, "Request should have been sent despite error response")
+    }
+
+    func test_performEventRequest_serverError_doesNotPropagateError() async {
+        StubURLProtocol.stubbedStatusCode = 500
+
+        await sut.performEventRequest(eventName, sessionID: sessionID)
+
+        XCTAssertNotNil(StubURLProtocol.lastRequest)
+    }
+
+    // MARK: - sendAnalyticsEvent
+
+    func test_sendAnalyticsEvent_dispatchesNetworkRequestInBackground() async {
+        StubURLProtocol.stubbedStatusCode = 200
+        let expectation = expectation(description: "Stub received the analytics request")
+        StubURLProtocol.onRequest = { _ in expectation.fulfill() }
+
+        sut.sendAnalyticsEvent(eventName, sessionID: sessionID)
+
+        await fulfillment(of: [expectation], timeout: 10.0)
+        XCTAssertEqual(StubURLProtocol.lastRequest?.url, analyticsURL)
+    }
+}
--- a/IntegrationTests/Info.plist
+++ b/IntegrationTests/Info.plist
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>en</string>
+	<key>CFBundleExecutable</key>
+	<string>$(EXECUTABLE_NAME)</string>
+	<key>CFBundleIdentifier</key>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleName</key>
+	<string>$(PRODUCT_NAME)</string>
+	<key>CFBundlePackageType</key>
+	<string>BNDL</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
+	<key>CFBundleVersion</key>
+	<string>1</string>
+</dict>
+</plist>
--- a/IntegrationTests/Mocks/MockScriptMessage.swift
+++ b/IntegrationTests/Mocks/MockScriptMessage.swift
@@ -1,0 +1,17 @@
+import WebKit
+
+final class MockScriptMessage: WKScriptMessage {
+
+    var stubName: String = ""
+    var stubBody: Any = [:]
+
+    override var name: String {
+        get { stubName }
+        set { stubName = newValue }
+    }
+
+    override var body: Any {
+        get { stubBody }
+        set { stubBody = newValue }
+    }
+}
--- a/IntegrationTests/Mocks/MockWebAuthSession.swift
+++ b/IntegrationTests/Mocks/MockWebAuthSession.swift
@@ -1,0 +1,18 @@
+import AuthenticationServices
+@testable import PopupBridge
+
+final class MockWebAuthSession: WebAuthenticationSession {
+
+    var startWasCalled = false
+    var capturedURL: URL?
+
+    override func start(
+        url: URL,
+        context: ASWebAuthenticationPresentationContextProviding,
+        sessionDidComplete: @escaping (URL?, Error?) -> Void,
+        sessionDidCancel: @escaping () -> Void
+    ) {
+        startWasCalled = true
+        capturedURL = url
+    }
+}
--- a/IntegrationTests/Mocks/StubURLProtocol.swift
+++ b/IntegrationTests/Mocks/StubURLProtocol.swift
@@ -1,0 +1,47 @@
+import Foundation
+
+final class StubURLProtocol: URLProtocol {
+
+    static var stubbedStatusCode: Int = 200
+    static var lastRequest: URLRequest?
+    static var onRequest: ((URLRequest) -> Void)?
+
+    static func reset() {
+        stubbedStatusCode = 200
+        lastRequest = nil
+        onRequest = nil
+    }
+
+    override class func canInit(with request: URLRequest) -> Bool { true }
+    override class func canonicalRequest(for request: URLRequest) -> URLRequest { request }
+
+    override func startLoading() {
+        var capturedRequest = request
+        if capturedRequest.httpBody == nil, let stream = capturedRequest.httpBodyStream {
+            var data = Data()
+            stream.open()
+            let buffer = UnsafeMutablePointer<UInt8>.allocate(capacity: 4096)
+            while stream.hasBytesAvailable {
+                let count = stream.read(buffer, maxLength: 4096)
+                if count > 0 { data.append(buffer, count: count) }
+            }
+            buffer.deallocate()
+            stream.close()
+            capturedRequest.httpBody = data
+        }
+        Self.lastRequest = capturedRequest
+        Self.onRequest?(capturedRequest)
+        guard let url = request.url,
+              let httpResponse = HTTPURLResponse(
+                url: url,
+                statusCode: Self.stubbedStatusCode,
+                httpVersion: nil,
+                headerFields: nil
+              ) else { return }
+        client?.urlProtocol(self, didReceive: httpResponse, cacheStoragePolicy: .notAllowed)
+        client?.urlProtocol(self, didLoad: Data())
+        client?.urlProtocolDidFinishLoading(self)
+    }
+
+    override func stopLoading() {}
+}
--- a/IntegrationTests/PopupBridge_IntegrationTests.swift
+++ b/IntegrationTests/PopupBridge_IntegrationTests.swift
@@ -1,0 +1,145 @@
+import AuthenticationServices
+import XCTest
+import WebKit
+@testable import PopupBridge
+
+final class PopupBridge_IntegrationTests: XCTestCase {
+
+    override func setUp() {
+        super.setUp()
+        StubURLProtocol.reset()
+        POPPopupBridge.analyticsService = AnalyticsService()
+    }
+
+    override func tearDown() {
+        StubURLProtocol.reset()
+        POPPopupBridge.analyticsService = AnalyticsService()
+        super.tearDown()
+    }
+
+    // MARK: - init
+
+    func test_init_withRealWKWebView_injectsUserScriptAtDocumentStart() {
+        let webView = WKWebView()
+        XCTAssertEqual(webView.configuration.userContentController.userScripts.count, 0)
+
+        let _ = POPPopupBridge(webView: webView)
+
+        XCTAssertEqual(webView.configuration.userContentController.userScripts.count, 1)
+        XCTAssertEqual(webView.configuration.userContentController.userScripts[0].injectionTime, .atDocumentStart)
+        XCTAssertFalse(webView.configuration.userContentController.userScripts[0].isForMainFrameOnly)
+    }
+
+    @MainActor
+    func test_init_withRealAnalyticsService_sendsStartedEvent() async {
+        let config = URLSessionConfiguration.ephemeral
+        config.protocolClasses = [StubURLProtocol.self]
+        POPPopupBridge.analyticsService = AnalyticsService(session: URLSession(configuration: config))
+
+        await withCheckedContinuation { (continuation: CheckedContinuation<Void, Never>) in
+            var resumed = false
+            StubURLProtocol.onRequest = { request in
+                guard let body = request.httpBody,
+                      let decoded = try? JSONDecoder().decode(FPTIBatchData.self, from: body),
+                      decoded.events.first?.fptiEvents.first?.eventName == PopupBridgeAnalytics.started,
+                      !resumed else { return }
+                resumed = true
+                continuation.resume()
+            }
+            let _ = POPPopupBridge(webView: WKWebView())
+        }
+    }
+
+    // MARK: - userContentController
+
+    func test_userContentController_withNonJSONSerializableBody_returnsEarlyWithoutStartingSession() {
+        let webView = WKWebView()
+        let mockSession = MockWebAuthSession()
+        let bridge = POPPopupBridge(webView: webView, webAuthenticationSession: mockSession)
+
+        let message = MockScriptMessage()
+        message.stubName = "POPPopupBridge"
+        message.stubBody = ["this-is-an-array-not-a-dict"]
+
+        bridge.userContentController(WKUserContentController(), didReceive: message)
+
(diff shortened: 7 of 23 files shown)


Sources/PopupBridge/Analytics/PopupBridgeAnalytics.swift
```diff
@@ -1,8 +1,12 @@
 enum PopupBridgeAnalytics {
-    
     static let started = "popup-bridge:started"
     static let succeeded = "popup-bridge:succeeded"
     static let failed = "popup-bridge:failed"
     static let canceled = "popup-bridge:canceled"
+    static let paypalInstalled = "popup-bridge:app-detection:paypal-installed"
+    static let paypalNotInstalled = "popup-bridge:app-detection:paypal-not-installed"
```
jaxdesmarais: This should live in a tag not on the events

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -16,27 +16,34 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
     ///   - prefersEphemeralWebBrowserSession: A Boolean that, when true, requests that the browser does not share cookies
-    ///   or other browsing data between the authenthication session and the user’s normal browser session.
+    ///   or other browsing data between the authenthication session and the user's normal browser session.
     ///   Defaults to `true`.
-    public init(webView: WKWebView, prefersEphemeralWebBrowserSession: Bool = true) {
+    ///   - enablePopupBridgeAppSwitch: When true, allows the SDK to launch the native PayPal app for checkout
+    ///   instead of opening a browser. Defaults to false for backward compatibility.
+    ///   This is specific to the popup bridge flow and is separate from the JS SDK's
+    ///   appSwitchWhenAvailable which controls non-webview mobile browser app switch.
+    public init(webView: WKWebView, prefersEphemeralWebBrowserSession: Bool = true, enablePopupBridgeAppSwitch: Bool = false) {
```
jaxdesmarais, line 38: why does this take a different approach vs venmo?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +62,118 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+                return
+            }
+
+            self?.injectWebView(webView: webView, withJavaScript: script)
+            return
+        }
+    }
+
     deinit {
+        if let observer = launchAppReturnObserver {
```
jaxdesmarais: we can use the existing name and pass it into the remove method:
```suggestion
        if let launchAppReturnObserver {
```

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +62,118 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
         webView.configuration.userContentController.removeAllScriptMessageHandlers()
     }
     
+    // MARK: - Launch App Return Handling
+
+    /// Starts listening for the return URL notification posted by the host app's SceneDelegate.
+    /// Called only after a successful launchApp; removed after handling.
+    private func startObservingLaunchAppReturn() {
+        // Remove any stale observer before adding a new one
+        if let observer = launchAppReturnObserver {
```
jaxdesmarais: Same here:

```suggestion
        if let launchAppReturnObserver {
```

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +62,118 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            NotificationCenter.default.removeObserver(observer)
+        }
+
+        launchAppReturnObserver = NotificationCenter.default.addObserver(
+            forName: Notification.Name("popupBridgeReturnURL"),
+            object: nil,
+            queue: .main
+        ) { [weak self] notification in
+            guard let self,
+                  let url = notification.userInfo?["url"] as? URL else {
```
```suggestion
            guard let self, let url = notification.userInfo?["url"] as? URL else {
```

jaxdesmarais: What if self is nil here?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +62,118 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            self.handleLaunchAppReturn(url: url)
+        }
+    }
+
+    /// Parses the return URL and injects `window.popupBridge.onComplete()` into the WebView.
+    /// Venice may return fragment-based data for web_sdk integration:
+    ///   merchantapp://popupbridgev1/onApprove#onApprove&PayerID=XXX&token=EC-YYY
+    /// The JS SDK is responsible for normalizing `path`, `queryItems`, and `hash`.
+    private func handleLaunchAppReturn(url: URL) {
+
```
```suggestion
```

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +62,118 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            guard let self,
+                  let url = notification.userInfo?["url"] as? URL else {
+                return
+            }
+            self.handleLaunchAppReturn(url: url)
+        }
+    }
+
+    /// Parses the return URL and injects `window.popupBridge.onComplete()` into the WebView.
+    /// Venice may return fragment-based data for web_sdk integration:
```
jaxdesmarais: Remove internal product name

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +62,118 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+    ///   merchantapp://popupbridgev1/onApprove#onApprove&PayerID=XXX&token=EC-YYY
+    /// The JS SDK is responsible for normalizing `path`, `queryItems`, and `hash`.
```
jaxdesmarais: Which JS SDK?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +62,118 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+    private func handleLaunchAppReturn(url: URL) {
+
+        // One-shot: stop observing immediately
```
jaxdesmarais: What does "one-shot" mean?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +62,118 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+        if let observer = launchAppReturnObserver {
```
```suggestion
        if let launchAppReturnObserver {
```

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +62,118 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+        } ?? [:]
+
+        let payload = URLDetailsPayload(
+            path: urlComponents.path,
+            queryItems: queryItems,
+            hash: urlComponents.fragment
+        )
+
+        if let payloadData = try? JSONEncoder().encode(payload),
+           let payloadString = String(data: payloadData, encoding: .utf8) {
```
jaxdesmarais: Either single line or newline after if

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -106,11 +220,34 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
             name: messageHandlerName
         )
         
+        let isPayPalInstalled = enablePopupBridgeAppSwitch && application.isPayPalAppInstalled()
+        if enablePopupBridgeAppSwitch {
+            Self.analyticsService.sendAnalyticsEvent(
+                isPayPalInstalled ? PopupBridgeAnalytics.paypalInstalled : PopupBridgeAnalytics.paypalNotInstalled,
```
jaxdesmarais: This should be a tag so it can be set on all events vs an independent event. Event names are for conversion only and everything else should go in a tag to avoid branching from event names.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -141,8 +278,35 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
                   let script = try? JSONDecoder().decode(WebViewMessage.self, from: jsonData) else {
                 return
             }
-            
-            if let urlString = script.url, let url = URL(string: urlString) {
+
+            if let launchAppURLString = script.launchApp, let launchAppURL = URL(string: launchAppURLString) {
+                Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.appLaunchStarted, sessionID: sessionID)
```
jaxdesmarais, line 294: How will this be used for conversion? Is there a different way to send this data so we can more accurately track success/failure of the app switch?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -141,8 +278,35 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+                application.openURL(launchAppURL) { [weak self] success in
+                    guard let self else { return }
+                    if success {
```
jaxdesmarais, line 298:
```suggestion
                    guard let self else { return }

                    if success {
```

Sources/PopupBridge/PopupBridgeUserScript.swift
```diff
@@ -1,40 +1,67 @@
...
 
 struct PopupBridgeUserScript {
-    
     let scheme: String
     let scriptMessageHandlerName: String
     let host: String
     let isVenmoInstalled: Bool
-    
+    let isPayPalInstalled: Bool
+    let returnUrlScheme: String?
```
jaxdesmarais:
```suggestion
    let returnURLScheme: String?
```

Sources/PopupBridge/PopupBridgeUserScript.swift
```diff
@@ -1,40 +1,67 @@
...
+
     var rawJavascript: String {
-        """
+        let returnUrlPrefix = "\(scheme)://\(host)/"
+        let deepLinkJs: String
```
jaxdesmarais:
```suggestion
        let returnURLPrefix = "\(scheme)://\(host)/"
        let deepLinkJS: String
```

Sources/PopupBridge/PopupBridgeUserScript.swift
```diff
@@ -1,40 +1,67 @@
...
+        let deepLinkProperty: String
+
+        if let returnUrlScheme {
+            let deepLinkReturnUrlPrefix = "\(returnUrlScheme)://\(host)/"
+            deepLinkJs = """
```
jaxdesmarais: What does deepLinkJs mean/do?

Sources/PopupBridge/PopupBridgeUserScript.swift
```diff
@@ -1,40 +1,67 @@
...
+
+                        window.popupBridge.getDeepLinkReturnUrlPrefix = function getDeepLinkReturnUrlPrefix() {
+                            return '\(deepLinkReturnUrlPrefix)';
+                        };
+            """
+            deepLinkProperty = """
```
jaxdesmarais, line 24: What does deepLinkProperty mean/do?

Sources/PopupBridge/UIApplication+URLOpener.swift
```diff
@@ -1,17 +1,29 @@
...
     func isVenmoAppInstalled() -> Bool {
         guard let venmoURL = URL(string: "com.venmo.touch.v2://") else {
             return false
         }
         return canOpenURL(venmoURL)
     }
+
+    /// Indicates whether the PayPal App is installed.
+    func isPayPalAppInstalled() -> Bool {
+        guard let paypalURL = URL(string: "paypal-app-switch-checkout://") else {
```
jaxdesmarais: iirc we should be using "paypal" in newer versions. This scheme existed in earlier versions of the PayPal app so we could differentiate which PayPal app versions supported app switch.

UnitTests/MockAnalyticsService.swift
```diff
@@ -6,10 +6,12 @@ class MockAnalyticsService: AnalyticsServiceable {
     var lastEventName: String?
     var lastSessionID: String?
     var eventCount = 0
+    var allEventNames: [String] = []
```
jaxdesmarais: what's this being used for?

UnitTests/MockURLOpener.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+@testable import PopupBridge
+
+class MockURLOpener: URLOpener {
+
+    var venmoInstalled = false
+    var paypalInstalled = false
```
jaxdesmarais: ensure we are using this casing across this PR:
```suggestion
    var payPalInstalled = false
```

Sources/PopupBridge/POPPopupBridge.swift
jtanya17: This PR follows the existing Popup Bridge pattern and only adds PayPal-specific app-switch handling behind the feature flag. I kept Venmo unchanged here to avoid expanding scope, since the HLD and QA for this work were focused on the PayPal path.

Sources/PopupBridge/POPPopupBridge.swift
jtanya17: Hey Jax, if self is nil, the bridge has already been deallocated and there’s nothing left to handle, so returning early is the correct behavior. I split the guards to make that explicit.

Sources/PopupBridge/POPPopupBridge.swift
jtanya17: I removed that wording and made the behavior explicit in code/comments instead.

Sources/PopupBridge/POPPopupBridge.swift
jtanya17: Agreed. I removed the PayPal app-detection analytics from this PR rather than keeping the wrong event shape.

Sources/PopupBridge/POPPopupBridge.swift
jtanya17: This one is intended to mark the launch attempt itself, separate from the eventual success/failure path. I left the existing `appLaunch*` events in place for now, but I agree the broader analytics model may need a follow-up if we want tighter conversion semantics.

Sources/PopupBridge/UIApplication+URLOpener.swift
jtanya17: That's a good callout. I updated the detection scheme to `paypal://` in the branch.

UnitTests/MockAnalyticsService.swift
jtanya17: It was only there to support the added analytics assertions. Since I removed the app-detection analytics tests, that helper is no longer needed for those cases.

UnitTests/MockURLOpener.swift
jtanya17: Updated!

Sources/PopupBridge/PopupBridgeUserScript.swift
jtanya17: `deepLinkJs` injects the `getDeepLinkReturnUrlPrefix()` JavaScript function onto `window.popupBridge`

Sources/PopupBridge/PopupBridgeUserScript.swift
jtanya17: `deepLinkProperty` injects the `deepLinkReturnUrlPrefix` JavaScript property onto `window.popupBridge`

Sources/PopupBridge/POPPopupBridge.swift
jtanya17: Removed

Sources/PopupBridge/POPPopupBridge.swift
jtanya17: Made the change

Sources/PopupBridge/POPPopupBridge.swift
jtanya17: The `if let launchAppReturnObserver` shorthand binding creates a new let constant that shadows the property, making it impossible to assign nil back. Renaming the binding to observer keeps the original `launchAppReturnObserver` property in scope and writable.

NastassiaRodzik (approved):

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +63,119 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            queryItems: queryItems,
+            hash: urlComponents.fragment
+        )
+
+        if let payloadData = try? JSONEncoder().encode(payload),
+            let payloadString = String(data: payloadData, encoding: .utf8) {
+            Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.succeeded, sessionID: sessionID)
+
+            let script = """
+                ;(function() {
```
buzzamus, line 175: I am assuming there is a specific reason this starts with a semicolon, but I am just curious what the reason is

buzzamus (changes requested):
It doesn't appear this PR builds. Looks like there are issues around `launchAppReturnObserver`. Once this builds and has been thoroughly tested please let us know when this is ready for review

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,12 +1,25 @@
...
- - [ ] Added a changelog entry
+### Checklist
+- [ ] Added a changelog entry
+- [ ] Tested and confirmed payment flows affected by this change are functioning as expected
 
 ### Authors
 > List GitHub usernames for everyone who contributed to this pull request.
-
-- 
+-
```
buzzamus, line 25: I don't think this file is from this PR. I see several files committed from previous PRs, so I think your git history is screwed up somehow.

Demo/Demo.xcodeproj/xcshareddata/xcschemes/UITests.xcscheme
```diff
@@ -10,7 +10,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      codeCoverageEnabled = "YES">
       <Testables>
          <TestableReference
             skipped = "NO">
```
buzzamus, line 17: Same here. I don't think this change is meant for this PR as it was already merged into main

IntegrationTests/Mocks/MockScriptMessage.swift
```diff
@@ -0,0 +1,17 @@
...
+    override var name: String {
+        get { stubName }
+        set { stubName = newValue }
+    }
+
+    override var body: Any {
+        get { stubBody }
+        set { stubBody = newValue }
+    }
+}
```
buzzamus, line 17: Same with this file. This is already merged into main

IntegrationTests/Mocks/MockWebAuthSession.swift
```diff
@@ -0,0 +1,18 @@
...
+    override func start(
+        url: URL,
+        context: ASWebAuthenticationPresentationContextProviding,
+        sessionDidComplete: @escaping (URL?, Error?) -> Void,
+        sessionDidCancel: @escaping () -> Void
+    ) {
+        startWasCalled = true
+        capturedURL = url
+    }
+}
```
buzzamus, line 18: Same here

IntegrationTests/Mocks/StubURLProtocol.swift
```diff
@@ -0,0 +1,47 @@
...
+                httpVersion: nil,
+                headerFields: nil
+              ) else { return }
+        client?.urlProtocol(self, didReceive: httpResponse, cacheStoragePolicy: .notAllowed)
+        client?.urlProtocol(self, didLoad: Data())
+        client?.urlProtocolDidFinishLoading(self)
+    }
+
+    override func stopLoading() {}
+}
```
buzzamus, line 47: Same here

IntegrationTests/AnalyticsService_IntegrationTests.swift
```diff
@@ -0,0 +1,74 @@
+import XCTest
```
buzzamus, line 1: This file also exists on main already from a previous PR

IntegrationTests/Info.plist
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
buzzamus, line 1: Same here

IntegrationTests/PopupBridge_IntegrationTests.swift
```diff
@@ -0,0 +1,145 @@
+import AuthenticationServices
```
buzzamus, line 1: Same here

PopupBridge.xcodeproj/xcshareddata/xcschemes/IntegrationTests.xcscheme
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
buzzamus, line 1: Same here

PopupBridge.xcodeproj/project.pbxproj
```diff
@@ -7,9 +7,13 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		085C05B32F9A78E000460CD9 /* MockWebAuthSession.swift in Sources */ = {isa = PBXBuildFile; fileRef = 085C05B22F9A78E000460CD9 /* MockWebAuthSession.swift */; };
+		085C05B82F9A79DA00460CD9 /* MockScriptMessage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 085C05B72F9A79DA00460CD9 /* MockScriptMessage.swift */; };
+		085C05BA2F9A7AB900460CD9 /* StubURLProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 085C05B92F9A7AB900460CD9 /* StubURLProtocol.swift */; };
 		451348272D8CAD7C009265C9 /* WebViewScriptHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 451348262D8CAD6D009265C9 /* WebViewScriptHandler.swift */; };
```
buzzamus, line 13: This project file has all of the changes from recent PRs in it, so this will also need to be resolved. Not sure if this is from a bad merge or a rebase screwed up the history, but these need to be removed. The best way will be to instead of manually removing see if you can fix the git history

SECURITY.md
```diff
@@ -0,0 +1,55 @@
...
+- Keep you informed of our progress toward a fix
+- Notify you before any public disclosure
+
+We ask that you:
+
+- Do not publicly disclose the issue before it has been resolved
+- Avoid accessing, modifying, or deleting data that does not belong to you
+- Make a good faith effort to avoid disruption to production systems
+
+We appreciate responsible disclosure and your efforts to keep Braintree SDK users safe.
```
buzzamus, line 55: This is another previously merged change

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +63,134 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
         self.init(webView: webView)
         self.webAuthenticationSession = webAuthenticationSession
     }
-    
+
+    /// Exposed for testing
+    convenience init(
+        webView: WKWebView,
```
buzzamus, line 69: There are quite a few initializers now in this file, which to me is a big code smell. I think we should try and find a different way to handle. If we are using default values this wouldn't be a breaking change so, I am not sure if we need so many new ones created

buzzamus (changes requested):
I think the git history got messed up somewhere since it has previous PR changes as part of this PR's commit history and also left a few other comments.

jaxdesmarais:
@jtanya17 I am moving this PR to a draft state based on @buzzamus comments as this code is not ready to be reviewed by my team. Please resolve our issues and take a look through your diff before reopening.

jtanya17:
Closing in favor of https://github.com/braintree/popup-bridge-ios/pull/95

PR #95, opened 2026-05-13
Enhance PopupBridge app switch flow with custom return URL and fallback validation

joaaraujo-paypal:
### Summary of changes

- Add PayPal app switch support to PopupBridge
- Introduces an optional `enablePayPalAppSwitch ` flag to `POPPopupBridge` that, when enabled, allows the SDK to launch the native PayPal app for checkout instead of opening a browser session. 
- When the app launch succeeds, a `NotificationCenter` observer handles the deep link return URL and injects the result back into the WebView. If the app launch fails, the flow falls back to the existing `WebAuthenticationSession` behavior. 
- Exposes `isPayPalInstalled`, `launchApp()`, and `getDeepLinkReturnUrlPrefix()` to JavaScript, and fires analytics events for app detection and launch outcomes.

Based on this PR: 
https://github.com/braintree/popup-bridge-ios/pull/90

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**

**Estimated AI Code Contribution**
- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry
- [x] Tested and confirmed payment flows affected by this change are functioning as expected

- @jtanya17 
- @joaaraujo-paypal

----
### Inner Source Process
Internal to PayPal contributors should fill out this section. All others can delete.

PR should follow these steps before codeowners review will begin:
1. PR should be opened in a draft state with the `tech lead review required`, and `inner source` label
2. PR should be reviewed by and approved by your team's technical lead, we do not allow LGTM reviews, there should be comments and feedback provided on all PR reviews
3. Once the above steps are completed, the PR can be moved to ready to review with the `tech lead review required` label removed
4. PR comments must be addressed within 24 hours, if you are unable to address within this timeframe, move the PR back to a draft state so our team knows not to review

### Inner Source Checklist
- [x] Added all labels to the PR
- [x] Provide steps to test the flows changed, if applicable in the summary
- [x] Demo video of the functionality, if applicable
- [x] All upstream dependencies are merged in and this PR can be released at any time; PRs should not be opened until this is true
- [x] Unit tests and builds have been run locally and pass/compile as expected


saralvasquez:
Please have your tech lead review this PR. Once it's approved by them you can remove the tech lead review required label and that will inform us that it's time for our team's review. Thanks!

vkspune (approved):

buzzamus (changes requested):
I think you need to rebase against main. There are 42 commits showing in the git history for this PR, but it looks like only about 6 or so are part of this PR's work

joaaraujo-paypal:
> Please have your tech lead review this PR. Once it's approved by them you can remove the tech lead review required label and that will inform us that it's time for our team's review. Thanks!

@vkspune reviewed this PR, but I'm not able to remove the tag of tech lead review

Sources/PopupBridge/Analytics/PopupBridgeAnalytics.swift
```diff
@@ -1,8 +1,9 @@
 enum PopupBridgeAnalytics {
-    
     static let started = "popup-bridge:started"
     static let succeeded = "popup-bridge:succeeded"
     static let failed = "popup-bridge:failed"
     static let canceled = "popup-bridge:canceled"
+    static let appLaunchStarted = "popup-bridge:app-launch:started"
+    static let appLaunchSucceeded = "popup-bridge:app-launch:succeeded"
+    static let appLaunchFailed = "popup-bridge:app-launch:failed"
```
jaxdesmarais:
```suggestion
    static let appLaunchStarted = "popup-bridge:app-switch:started"
    static let appLaunchSucceeded = "popup-bridge:app-switch:succeeded"
    static let appLaunchFailed = "popup-bridge:app-switch:failed"
```

Sources/PopupBridge/Analytics/PopupBridgeAnalytics.swift
```diff
@@ -1,8 +1,9 @@
 enum PopupBridgeAnalytics {
-
```
jaxdesmarais, line 2: Add back linebreaks in this file here and at bottom

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -16,27 +16,56 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
     private let hostName = "popupbridgev1"
     private let sessionID = UUID().uuidString.replacingOccurrences(of: "-", with: "")
     private let webView: WKWebView
-    private let application: URLOpener = UIApplication.shared
+    private let application: URLOpener
```
jaxdesmarais: Why this change? We should be using the singleton.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -16,27 +16,56 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+    ///           object: nil,
+    ///           userInfo: ["url": url]
+    ///       )
+    ///   }
+    ///   ```
+    ///   - returnURLScheme: The URL scheme registered in the app's `Info.plist` that PopupBridge should
+    ///   use as the return URL for the checkout flow. If `nil`, PopupBridge will attempt to read the first
+    ///   scheme from `CFBundleURLTypes`. Providing this value explicitly is recommended when the app
+    ///   registers multiple URL schemes (e.g. Facebook, Google Sign-In).
+    public init(webView: WKWebView, prefersEphemeralWebBrowserSession: Bool = true, enablePopupBridgeAppSwitch: Bool = false, returnURLScheme: String? = nil) {
```
jaxdesmarais: needs to be explicit, we also app switch for Venmo.
```suggestion
    public init(webView: WKWebView, prefersEphemeralWebBrowserSession: Bool = true, enablePayPalAppSwitch: Bool = false, returnURLScheme: String? = nil) {
```

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -16,27 +16,56 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
         self.webView = webView
-        
+        self.application = UIApplication.shared
```
jaxdesmarais: Let's revert this change and set it at the top of the file as it was before.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +84,127 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            NotificationCenter.default.removeObserver(launchAppReturnObserver)
+        }
         webView.configuration.userContentController.removeAllScriptMessageHandlers()
     }
     
+    // MARK: - Launch App Return Handling
+
+    /// Starts listening for the return URL notification posted by the host app's SceneDelegate.
+    /// Called only after a successful launchApp; removed after handling.
+    private func startObservingLaunchAppReturn() {
```
jaxdesmarais: Needs to be explicit to PayPal in naming, we support app switch with Venmo already. Is there a reason why the PayPal app switch flow is so much more complex vs Venmo?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +84,127 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+        if let launchAppReturnObserver {
+            NotificationCenter.default.removeObserver(launchAppReturnObserver)
+        }
+
+        launchAppReturnObserver = NotificationCenter.default.addObserver(
+            forName: Notification.Name(PopupBridgeConstants.notificationName),
+            object: nil,
+            queue: .main
```
jaxdesmarais: What's the ramification of this if the merchant backgrounds the merchant or PayPal app if any?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +84,127 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            self.handleLaunchAppReturn(url: url)
+        }
+    }
+
+    /// Parses the return URL and injects `window.popupBridge.onComplete()` into the WebView.
+    /// The return URL may include fragment-based data:
+    ///   merchantapp://popupbridgev1/onApprove#onApprove&PayerID=XXX&token=EC-YYY
+    /// Popup Bridge JavaScript is responsible for normalizing `path`, `queryItems`, and `hash`.
+    private func handleLaunchAppReturn(url: URL) {
+
```
jaxdesmarais: nit: remove extra linebreak

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +84,127 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+        // One-shot: stop observing immediately
```
jaxdesmarais: What does this mean?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +84,127 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            path: urlComponents.path,
+            queryItems: queryItems,
+            hash: urlComponents.fragment
+        )
+
+        if let payloadData = try? JSONEncoder().encode(payload),
+            let payloadString = String(data: payloadData, encoding: .utf8) {
+            Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.succeeded, sessionID: sessionID)
+
+            // The leading semicolon is a defensive IIFE guard: if the previous JS statement
```
jaxdesmarais: This is a helpful docstring - thanks!

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -86,11 +231,11 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
         )
 
         if let payloadData = try? JSONEncoder().encode(payload),
-           let payload = String(data: payloadData, encoding: .utf8) {
+            let payload = String(data: payloadData, encoding: .utf8) {
             Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.succeeded, sessionID: sessionID)
             return "window.popupBridge.onComplete(null, \(payload));"
         } else {
-            let errorMessage = "Failed to parse query items from return URL."
+            let errorMessage = "Failed to serialize return URL payload as JSON."
```
jaxdesmarais: Lets not change existing error messages

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -106,11 +251,35 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
             name: messageHandlerName
         )
         
+        // Even if the PayPal app is physically installed on the device, we intentionally
+        // report it as absent to the JavaScript layer when enablePopupBridgeAppSwitch is false.
+        // This ensures the JS SDK never attempts the native app switch path unless the
+        // integrator has explicitly opted in, preserving backward-compatible behavior.
+        let isPayPalInstalled = enablePopupBridgeAppSwitch && application.isPayPalAppInstalled()
```
jaxdesmarais: The naming of this is confusing since it's checking more than installed. Thoughts on something like `isPayPalAppSwitchEnabled` instead?

Sources/PopupBridge/PopupBridgeConstants.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-class PopupBridgeConstants {
-
```
jaxdesmarais, line 4: Lets keep this space since it's a general code style in our codebase

Sources/PopupBridge/PopupBridgeUserScript.swift
```diff
@@ -1,40 +1,66 @@
 import Foundation
 
 struct PopupBridgeUserScript {
-
```
jaxdesmarais, line 4: Add back space

Sources/PopupBridge/PopupBridgeUserScript.swift
```diff
@@ -1,40 +1,66 @@
...
     var rawJavascript: String {
-        """
+        let returnURLPrefix = "\(scheme)://\(host)/"
+        let deepLinkJS: String
+        let deepLinkProperty: String
+
+        if let returnURLScheme {
+            let deepLinkReturnURLPrefix = "\(returnURLScheme)://\(host)/"
+            deepLinkJS = """
+
```
jaxdesmarais: Is this space necessary for the JS?

Sources/PopupBridge/UIApplication+URLOpener.swift
```diff
@@ -1,17 +1,29 @@
 import UIKit
 
 protocol URLOpener {
-
```
jaxdesmarais, line 4: Add back space here and below

Sources/PopupBridge/WebViewMessage.swift
```diff
@@ -1,8 +1,8 @@
 /// A model type to represent details in the [`WKScriptMessage.body`](https://developer.apple.com/documentation/webkit/wkscriptmessage/1417901-body), sent by JavaScript code from a webpage.
 struct WebViewMessage: Codable {
-
```
jaxdesmarais, line 3: Add back space

Sources/PopupBridge/WebViewMessage.swift
```diff
@@ -1,8 +1,8 @@
 /// A model type to represent details in the [`WKScriptMessage.body`](https://developer.apple.com/documentation/webkit/wkscriptmessage/1417901-body), sent by JavaScript code from a webpage.
 struct WebViewMessage: Codable {
-    
     let url: String?
     let message: MessageDetails?
+    let launchApp: String?
```
jaxdesmarais:
```suggestion
    let launchPayPalAppSwitch: String?
```

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # PopupBridge iOS Release Notes
 
+## unreleased
+
+* Add optional `enablePopupBridgeAppSwitch` flag to `POPPopupBridge` initializer. When enabled, the SDK attempts to launch the native PayPal app for checkout instead of opening a browser session. Falls back to `ASWebAuthenticationSession` if the app launch fails.
+  * **Action required:** when `enablePopupBridgeAppSwitch: true` is set, the host app must implement `scene(_:openURLContexts:)` in its `SceneDelegate` and post a `NotificationCenter` notification named `PopupBridgeConstants.notificationName` with the return URL. Without this hook, the checkout flow will hang after the PayPal app returns. See the README for the required code snippet.
+* Expose `isPayPalInstalled`, `launchApp()`, and `getDeepLinkReturnUrlPrefix()` on `window.popupBridge` in JavaScript.
+* Add optional `returnURLScheme` parameter to `POPPopupBridge` initializer. Allows integrators with multiple URL schemes registered in `Info.plist` (e.g. Facebook, Google Sign-In) to explicitly specify which scheme PopupBridge should use as the return URL, avoiding ambiguity in the automatic `CFBundleURLTypes` lookup.
```
jaxdesmarais: This should only be the merchant facing changes. Anything implementation related should be in the readme. We should keep this limited to: add support for the PayPal app switch flow and the new public properties in the init.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -141,8 +310,42 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+                        let cancelScript = """
+                            if (typeof window.popupBridge.onCancel === 'function') {\
+                                window.popupBridge.onCancel();\
+                            } else {\
+                                window.popupBridge.onComplete(null, null);\
+                            }
+                        """
+                        let scheme = launchAppURL.scheme?.lowercased()
+                        guard scheme == "https" || scheme == "http" else {
+                            Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.canceled, sessionID: self.sessionID)
```
jaxdesmarais: This event seems incorrect, should be failed not canceled.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +84,127 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+
+            self.handleLaunchAppReturn(url: url)
+        }
+    }
+
+    /// Parses the return URL and injects `window.popupBridge.onComplete()` into the WebView.
+    /// The return URL may include fragment-based data:
+    ///   merchantapp://popupbridgev1/onApprove#onApprove&PayerID=XXX&token=EC-YYY
+    /// Popup Bridge JavaScript is responsible for normalizing `path`, `queryItems`, and `hash`.
+    private func handleLaunchAppReturn(url: URL) {
```
jaxdesmarais: Shouldn't we be doing some sort of validation here against the URLs? As this is written any code in the app that posts to `popupBridgeReturnURL` will cause arbitrary URL data to be injected into the webview.

jaxdesmarais: Also this entire method has no tests

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -106,11 +251,35 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            guard let urlTypes = Bundle.main.object(forInfoDictionaryKey: "CFBundleURLTypes") as? [[String: Any]] else {
+                return nil
+            }
+            for urlType in urlTypes {
+                if let schemes = urlType["CFBundleURLSchemes"] as? [String], let scheme = schemes.first {
+                    return scheme
+                }
+            }
+            return nil
+        }()
```
jaxdesmarais: This is really hard to reason about inside this method. We should be able to extract this into a private method.

Sources/PopupBridge/POPPopupBridge.swift
joaaraujo-paypal: We still use the `UIApplication.shared` singleton in production — it's just injected via `init (self.application = UIApplication.shared)` instead of being hardcoded on the property, so tests can pass a MockURLOpener.

Sources/PopupBridge/POPPopupBridge.swift
joaaraujo-paypal: Answered on last comment about it.

Sources/PopupBridge/POPPopupBridge.swift
joaaraujo-paypal: Added a PayPal naming here, and in other functions with same scenario. 
I don't know how to answer it, because we don't know flow of Venmo

Sources/PopupBridge/POPPopupBridge.swift
joaaraujo-paypal: It means the native side only does a raw split of the return URL into its components — path, queryItems, and hash — and hands them to the web layer as-is via window.popupBridge.onComplete(null, payload). We don't merge, decode, or interpret those fields here.

The actual normalization (e.g. combining query params with hash params, parsing the hash into key/values, building the final object the merchant's JS consumes) happens in the Popup Bridge JavaScript library on the page. The comment is just clarifying that division of responsibility so nobody adds URL-massaging logic on the native side that the JS already handles.

Sources/PopupBridge/WebViewMessage.swift
joaaraujo-paypal: This change will affect our contract with JS... So I did dis change but added a `CodingKeys` to keep our contract.

Sources/PopupBridge/PopupBridgeUserScript.swift
joaaraujo-paypal: Not necessary, removed.

CHANGELOG.md
joaaraujo-paypal: Updated Changelog and Readme

Sources/PopupBridge/Analytics/PopupBridgeAnalytics.swift
jaxdesmarais: Doesn't look like this was changed

Sources/PopupBridge/PopupBridgeConstants.swift
```diff
@@ -1,6 +1,8 @@
 import Foundation
 
-class PopupBridgeConstants {
+public class PopupBridgeConstants {
```
jaxdesmarais: This doesn't need to be public

Sources/PopupBridge/PopupBridgeConstants.swift
```diff
@@ -1,6 +1,8 @@
...
 
     static let callbackURLScheme: String = "sdk.ios.popup-bridge"
+    public static let notificationName: String = "popupBridgeReturnURL"
```
agedd: nit: i believe this notification name doesn't need to be public either

Sources/PopupBridge/Analytics/PopupBridgeAnalytics.swift
joaaraujo-paypal: Sorry, I miss it.

Sources/PopupBridge/PopupBridgeConstants.swift
joaaraujo-paypal: I think we need it public, because when enablePayPalAppSwitch is enabled, the SDK launches the native PayPal app for checkout and waits for the app to return via a deep link. For PopupBridge to receive that return URL, your SceneDelegate must post a NotificationCenter notification. Without this, the checkout flow will hang indefinitely after the PayPal app returns.
This is documented on Readme.

Sources/PopupBridge/PopupBridgeConstants.swift
joaaraujo-paypal: Just repeating the answer above:
I think we need it public, because when enablePayPalAppSwitch is enabled, the SDK launches the native PayPal app for checkout and waits for the app to return via a deep link. For PopupBridge to receive that return URL, your SceneDelegate must post a NotificationCenter notification. Without this, the checkout flow will hang indefinitely after the PayPal app returns.
This is documented on Readme.

Sources/PopupBridge/PopupBridgeConstants.swift
jaxdesmarais: This is very different than any other app switch flow including Venmo. We need to rethink this since this is code smell for merchants to have access to and need to do.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -141,8 +347,42 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
                   let script = try? JSONDecoder().decode(WebViewMessage.self, from: jsonData) else {
                 return
             }
-            
-            if let urlString = script.url, let url = URL(string: urlString) {
+
+            if let launchAppURLString = script.launchPayPalAppSwitch, let launchAppURL = URL(string: launchAppURLString) {
+                Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.appLaunchStarted, sessionID: sessionID)
+                application.openURL(launchAppURL) { [weak self] success in
```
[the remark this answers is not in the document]
joaaraujo-paypal: This is treated by launchPayPalAppSwitch. We dont need to check with `enablePayPalAppSwitch`. And the URL to be opened, it comes from backend.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +84,155 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            return false
+        }
+
+        // Scheme must match the return URL scheme PopupBridge advertised to PayPal. When a scheme
+        // can be resolved, enforce it so URLs minted for other schemes are rejected.
+        if let expectedScheme = resolveReturnURLScheme() {
+            return components.scheme?.caseInsensitiveCompare(expectedScheme) == .orderedSame
+        }
+
+        return true
```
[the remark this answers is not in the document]
joaaraujo-paypal: Fixed

Sources/PopupBridge/PopupBridgeConstants.swift
```diff
@@ -1,6 +1,8 @@
...
+
```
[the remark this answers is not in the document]
joaaraujo-paypal: Fixed

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -55,11 +84,155 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            """
+            injectWebView(webView: webView, withJavaScript: script)
+        } else {
+            let errorMessage = "Failed to serialize return URL payload as JSON."
+            let errorResponse = "new Error(\"\(errorMessage)\")"
+            Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.failed, sessionID: sessionID)
+            let script = "window.popupBridge.onComplete(\(errorResponse), null);"
+            injectWebView(webView: webView, withJavaScript: script)
+        }
+    }
```
buzzamus: This function is pretty long...is there any way we can break this up so that it is more readable?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -141,8 +347,42 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+                                returnBlock(url)
+                            }
+                        } sessionDidCancel: {
+                            Self.analyticsService.sendAnalyticsEvent(PopupBridgeAnalytics.canceled, sessionID: self.sessionID)
+                            self.injectWebView(webView: self.webView, withJavaScript: cancelScript)
+                        }
+                    }
+                }
+            } else if let urlString = script.url, let url = URL(string: urlString) {
                 webAuthenticationSession.start(url: url, context: self) { url, _ in
```
buzzamus: Same here. This method has now ballooned in size. This whole code base actually has increased with this PR and sort of sprinkles the app switch logic throughout.

buzzamus (changes requested):
What if we made a separate class to handle a lot of the app switch behavior? This could be held in a `AppSwitchSession` or something. Venmo doesn't actually have any sort of app switch logic within this repo and only sends a flag back to the JS script. 

I think if the strictly app switch logic was isolated it would require less drastic changes in the main code and PopupBridge continues to just be a coordinator.

joaaraujo-paypal:
> What if we made a separate class to handle a lot of the app switch behavior? This could be held in a `AppSwitchSession` or something. Venmo doesn't actually have any sort of app switch logic within this repo and only sends a flag back to the JS script.
> 
> I think if the strictly app switch logic was isolated it would require less drastic changes in the main code and PopupBridge continues to just be a coordinator.

We separated all files of PayPalAppSwitch, and let PopupBridge file be only a coordinator.

Sources/PopupBridge/PopupBridgeConstants.swift
```diff
@@ -1,6 +1,9 @@
 import Foundation
 
-class PopupBridgeConstants {
+public class PopupBridgeConstants {
 
     static let callbackURLScheme: String = "sdk.ios.popup-bridge"
+    static let host: String = "popupbridgev1"
+    public static let notificationName = Notification.Name("popupBridgeReturnURL")
```
jaxdesmarais: This makes the return-URL transport part of the public API. I'd remove it in favor of an explicit `POPPopupBridge.handleReturnURL(_:)` method (see top-level comment) so the transport stays internal and changeable. Additionally, we need to look into why we need all this additional logic for PayPal app switch but not venmo (@buzzamus already called this out)

Sources/PopupBridge/PayPalAppSwitchHandler.swift
```diff
@@ -0,0 +1,214 @@
...
+                self.onLaunchFailed(url)
+            }
+        }
+    }
+
+    // MARK: - Return Handling
+
+    /// Starts listening for the return URL notification posted by the host app's SceneDelegate.
+    /// Called only after a successful launch; removed after handling.
+    private func startObserving() {
```
jaxdesmarais: This global observer is the crux of the antipattern: any code in the app can post here, the `userInfo["url"]` contract is untyped, and a missing/incorrect post hangs the flow with no signal. Replace with a "pending 
handler" registration on successful launch, dispatched to by an explicit `handleReturnURL(_:)` entry point  same routing model as `BTAppContextSwitcher.handleOpen(_:)`.

Sources/PopupBridge/PayPalAppSwitchHandler.swift
```diff
@@ -0,0 +1,214 @@
...
+    ///
+    /// Parses the return URL and produces the `window.popupBridge.onComplete()` JavaScript for the
+    /// WebView. The return URL may include fragment-based data:
+    ///   merchantapp://popupbridgev1/onApprove#onApprove&PayerID=XXX&token=EC-YYY
+    /// Popup Bridge JavaScript is responsible for normalizing `path`, `queryItems`, and `hash`.
+    func handlePayPalLaunchAppReturn(url: URL) {
+        // Validate before doing anything: the return notification is an open channel, so any code
+        // in the host app could post an arbitrary URL to it. Only PopupBridge return URLs may be
+        // injected into the WebView. Invalid posts are ignored without consuming the one-shot
+        // observer, so a later legitimate return still completes the flow.
```
jaxdesmarais: This comment is the tell that the transport is the problem, we're hardening against arbitrary in-app posts. With an explicit handler the channel closes; keep `isValidPayPalReturnURL` but demote it to a normal routing check (`canHandleReturnURL`) rather than a security boundary.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -7,36 +7,67 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+    ///   In your `SceneDelegate`:
+    ///   ```swift
+    ///   func scene(_ scene: UIScene, openURLContexts URLContexts: Set<UIOpenURLContext>) {
+    ///       guard let url = URLContexts.first?.url else { return }
+    ///       NotificationCenter.default.post(
+    ///           name: PopupBridgeConstants.notificationName,
+    ///           object: nil,
+    ///           userInfo: ["url": url]
+    ///       )
+    ///   }
```
jaxdesmarais: Once we move to an explicit handler, update this doc to show `POPPopupBridge.handleReturnURL(url)`  instead of  the notification post. Also worth calling out the Bool return so integrators know they can chain to other app-switch handlers.

Sources/PopupBridge/PayPalAppSwitchHandler.swift
```diff
@@ -0,0 +1,214 @@
...
+        if let returnURLScheme {
+            return returnURLScheme
+        }
+
+        guard let urlTypes = Bundle.main.object(forInfoDictionaryKey: "CFBundleURLTypes") as? [[String: Any]] else {
+            return nil
+        }
+
+        for urlType in urlTypes {
+            if let schemes = urlType["CFBundleURLSchemes"] as? [String], let scheme = schemes.first {
```
jaxdesmarais: Separate/smaller: falling back to the first `CFBundleURLTypes` scheme is fragile for multi-scheme apps (as the comment notes). Consider requiring `returnURLScheme` in a new PayPal app switch init for the app-switch path rather than guessing.

jaxdesmarais (changes requested):
We need to reconsider the `SceneDelegate NotificationCenter` return-URL transport before this merges. The app-switch return is delivered by having merchants post to a public global notification (`PopupBridgeConstants.notificationName`) from their `SceneDelegate`, which the SDK observes. I'd push back on this transport (the rest of the PR: validation, JS injection, analytics, pendingResult handling looks good but this is an anti pattern and will lead to long term maintenence and security concerns). 

My concerns:
1. It's a global broadcast bus, so the SDK has to defend against itself. The code comment in `handlePayPalLaunchAppReturn` says it directly: "the return notification is an open channel, so any code in the host app could post an arbitrary URL to it." `isValidPayPalReturnURL` exists to patch a hole the design opens. An explicit API call has no open channel.                                                 
2. Strongly-typed, unenforced contract. Merchants must match a magic notification name and a `userInfo["url"]` key holding a URL (not String). None of it is in a signature; any mistake silently no-ops.
3. Silent hang on misintegration. The documented failure mode is "the checkout flow will hang indefinitely", no error, no timeout, no analytics. Rough for a payments flow.
4. Spooky action at a distance. Control flows merchant → global bus → deep inside the SDK with no visible call edge; hard to discover, debug, and unit-test.
5. Leaks internals into the public API. `public static let notificationName` freezes the transport into the public contract.
6. Inconsistent with our own ecosystem. `braintree_ios` solves the identical problem with `BTAppContextSwitcher.sharedInstance.handleOpenURL(context:)`. A merchant using both SDKs would call a method for Braintree but post a notification for PopupBridge.
7. No `objc` interoperability and exposing constants is an anti pattern

Sources/PopupBridge/PopupBridgeConstants.swift
buzzamus: Yeah, I am curious why Venmo just sends a notification if appSwitch is available or not and then just has the JS handle and why this isn't the same pattern we could take for PayPal app switch

joaaraujo-paypal:
We did the changes, removing notifications that could cause some leaks. And now using a pattern similar to BTSDK, creating a PayPalHandler to encapsulate all exclusive logic for PayPal.

buzzamus:
> We did the changes, removing notifications that could cause some leaks. And now using a pattern similar to BTSDK, creating a PayPalHandler to encapsulate all exclusive logic for PayPal.

@joaaraujo-paypal can you discuss this question further as well too please:

"Yeah, I am curious why Venmo just sends a notification if appSwitch is available or not and then just has the JS handle and why this isn't the same pattern we could take for PayPal app switch"

I am trying to understand the reason behind the differences so I can better review this change

Sources/PopupBridge/Analytics/PopupBridgeAnalytics.swift
```diff
@@ -1,8 +1,11 @@
 enum PopupBridgeAnalytics {
-    
+
     static let started = "popup-bridge:started"
     static let succeeded = "popup-bridge:succeeded"
     static let failed = "popup-bridge:failed"
     static let canceled = "popup-bridge:canceled"
-}
+    static let appLaunchStarted = "popup-bridge:app-switch:started"
```
agedd: niti: can we rename all the analytic events to match the string description. i.e.

`static let appSwitchStarted = "popup-bridge:app-switch:started"`

Sources/PopupBridge/PayPalAppSwitchHandler.swift
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    // MARK: - Coordinator Queries
+
+    /// The URL scheme used as the checkout return URL, advertised to the JS layer.
+    var resolvedReturnURLScheme: String? {
+        resolveReturnURLScheme()
+    }
+
+    /// Whether the PayPal app is installed and can be switched to.
```
agedd: nit:  can delete the doc-strings here, the purpose of the function is pretty clear

Sources/PopupBridge/PayPalAppSwitchHandler.swift
```diff
@@ -0,0 +1,174 @@
...
+    func isPayPalAppInstalled() -> Bool {
+        application.isPayPalAppInstalled()
+    }
+
+    // MARK: - Launch
+
+    /// Attempts a native PayPal app switch for the given URL. On success, registers as the pending
+    /// handler so the SceneDelegate return can be routed back via `PopupBridgeAppContextSwitcher`;
+    /// on failure, reports back so the coordinator can fall back or cancel.
+    func launch(url: URL) {
```
agedd: nit: inline the parameter here

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -7,36 +7,65 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+
+    private let enablePayPalAppSwitch: Bool
     private var webAuthenticationSession: WebAuthenticationSession = WebAuthenticationSession()
     private var returnBlock: ((URL) -> Void)? = nil
-    
+
+    /// Owns the strictly PayPal app switch flow (native launch + SceneDelegate return handling),
+    /// keeping `POPPopupBridge` a coordinator. Implicitly unwrapped because its callbacks capture
+    /// `self`, so it can only be created after `super.init()`.
+    private var appSwitchHandler: PayPalAppSwitchHandler!
```
agedd: nit: it is not advisable to force-unwrap, can we make this an optional instead

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -7,36 +7,65 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+    ///   func scene(_ scene: UIScene, openURLContexts URLContexts: Set<UIOpenURLContext>) {
+    ///       guard let url = URLContexts.first?.url else { return }
+    ///       PopupBridgeAppContextSwitcher.shared.handleReturnURL(url)
+    ///   }
+    ///   ```
+    ///   - returnURLScheme: The URL scheme registered in the app's `Info.plist` that PopupBridge should
+    ///   use as the return URL for the checkout flow. **Required when `enablePayPalAppSwitch` is `true`** —
+    ///   PopupBridge does not guess the scheme from `CFBundleURLTypes`, since apps that register multiple
+    ///   URL schemes (e.g. Facebook, Google Sign-In) would resolve the wrong one.
+    public init(webView: WKWebView, prefersEphemeralWebBrowserSession: Bool = true, enablePayPalAppSwitch: Bool = false, returnURLScheme: String? = nil) {
```
agedd: nit: can we newline each of these parameters

agedd: also let's add docstrings for this `init` method

joaaraujo-paypal:
> > We did the changes, removing notifications that could cause some leaks. And now using a pattern similar to BTSDK, creating a PayPalHandler to encapsulate all exclusive logic for PayPal.
> 
> @joaaraujo-paypal can you discuss this question further as well too please:
> 
> "Yeah, I am curious why Venmo just sends a notification if appSwitch is available or not and then just has the JS handle and why this isn't the same pattern we could take for PayPal app switch"
> 
> I am trying to understand the reason behind the differences so I can better review this change

@buzzamus  sorry I miss this answer.
The difference is what the native side actually does in each flow.

From what I saw, in **Venmo**, PopupBridge only job is to report whether the app is installed `isVenmoInstalled`, the JS SDK owns the rest. There's no native launch and no deep-link return coming back outside the WebView, so a one-way boolean is enough.

For PayPal app switch, the native side launches the PayPal app and receives the return as a deep link in the host app's `SceneDelegate` outside the WebView, with no JS context. That URL has to be routed back to the awaiting PopupBridge instance, validated, and injected to complete the flow. Only a flag can't do that.

That's why PayPal needs an explicit entry point `PopupBridgeAppContextSwitcher.shared.handleReturnURL(_:))`, mirroring BTAppContextSwitcher in braintree_ios. If Venmo ever gains a native launch + return here, it'd use the same mechanism.

Sources/PopupBridge/PayPalAppSwitchHandler.swift
joaaraujo-paypal: What do you mean `inline`? Just omit the parameter name like this:
```
func launch(_ url: URL) {
```
?

Sources/PopupBridge/POPPopupBridge.swift
joaaraujo-paypal: This `init` already has a docstrings, so I think you wanna say that in other inits we don't have it. So I've added now.

Sources/PopupBridge/PopupBridgeConstants.swift
jaxdesmarais: This class should not be exposed publicly

Sources/PopupBridge/POPPopupBridge.swift
jaxdesmarais: You can still do that, we don't need to change this pattern as it's doing the same thing. Make it a var with the default value and then you can set the property in the tests if needed. We do the same thing in the core SDK.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -7,36 +7,69 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+    ///   ```
+    ///   - returnURLScheme: The URL scheme registered in the app's `Info.plist` that PopupBridge should
+    ///   use as the return URL for the checkout flow. **Required when `enablePayPalAppSwitch` is `true`** —
+    ///   PopupBridge does not guess the scheme from `CFBundleURLTypes`, since apps that register multiple
+    ///   URL schemes (e.g. Facebook, Google Sign-In) would resolve the wrong one.
+    public init(
+        webView: WKWebView,
+        prefersEphemeralWebBrowserSession: Bool = true,
+        enablePayPalAppSwitch: Bool = false,
+        returnURLScheme: String? = nil
```
jaxdesmarais: The logic around this seems very brittle since if merchants configure the `returnURLScheme` incorrectly we will have to troubleshoot an incorrect integration or rely on the top most scene, which can lead to a poor user experience. We should instead have 2 init, 1 for app switch PayPal with `returnURLScheme` required. If we take that approach we can likely remove the `enablePayPalAppSwitch` all together. Alternatively you can keep both the `enablePayPalAppSwitch` and `returnURLScheme` as required in a second init.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -47,46 +80,121 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+    ///
+    /// Convenience initializer that injects both a custom `WebAuthenticationSession` and a `URLOpener`,
+    /// allowing tests to exercise the PayPal app switch path end to end with mocked dependencies.
+    /// - Parameters:
+    ///   - webView: The web view to add a script message handler to.
+    ///   - webAuthenticationSession: The (typically mocked) session used to drive the web auth flow.
+    ///   - enablePayPalAppSwitch: When true, enables the native PayPal app switch path. Defaults to false.
+    ///   - returnURLScheme: The return URL scheme advertised to PayPal. Required when `enablePayPalAppSwitch` is true.
+    ///   - application: The (typically mocked) `URLOpener` used to detect installed apps and open URLs.
+    convenience init(
```
jaxdesmarais: We can remove this entirely if we make `application` a var with default property. This is a much cleaner approach.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -47,46 +80,121 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+    /// Internal designated initializer that accepts a `URLOpener`, letting tests inject a mock instead
+    /// of `UIApplication.shared`. Mirrors the public `init` but with the injectable application.
+    /// - Parameters:
+    ///   - webView: The web view to add a script message handler to.
+    ///   - prefersEphemeralWebBrowserSession: When true, the browser does not share cookies or browsing
+    ///   data between the authentication session and the user's normal browser session. Defaults to `true`.
+    ///   - enablePayPalAppSwitch: When true, enables the native PayPal app switch path. Defaults to false.
+    ///   - returnURLScheme: The return URL scheme advertised to PayPal. Required when `enablePayPalAppSwitch` is true.
+    ///   - application: The `URLOpener` used to detect installed apps and open URLs.
+    init(
```
jaxdesmarais, line 126: Same here, we don't need this and is just adding more code than necessary.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -47,46 +80,121 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+    }
+
     deinit {
         webView.configuration.userContentController.removeAllScriptMessageHandlers()
     }
-    
+
+    /// Builds the app switch handler, wiring its results back to this coordinator: completion JS is
+    /// injected into the WebView, and a failed native launch falls back to a web authentication session.
+    private func makeAppSwitchHandler(application: URLOpener, returnURLScheme: String?) -> PayPalAppSwitchHandler {
```
jaxdesmarais: nit once we make `returnURLScheme` required:
```suggestion
    private func buildAppSwitchHandler(with application: URLOpener, returnURLScheme: String) -> PayPalAppSwitchHandler {
```

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -47,46 +80,121 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
 
-        let queryItems = urlComponents.queryItems?.reduce(into: [:]) { partialResult, queryItem in
-            partialResult[queryItem.name] = queryItem.value
-        }
-
-        let payload = URLDetailsPayload(
-            path: urlComponents.path,
-            queryItems: queryItems ?? [:],
-            hash: urlComponents.fragment
-        )
```
jaxdesmarais, line 141: Can you help me understand this deletion?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -105,20 +213,38 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
             WebViewScriptHandler(proxy: self),
             name: messageHandlerName
         )
-        
+
+        // The app switch path requires an explicit returnURLScheme. If the integrator opted into
+        // enablePayPalAppSwitch but didn't provide one, fail fast (loud in debug) and keep app
+        // switch disabled rather than guessing a scheme or hanging the flow at return time.
+        if enablePayPalAppSwitch && appSwitchHandler?.resolvedReturnURLScheme == nil {
+            assertionFailure("enablePayPalAppSwitch requires a returnURLScheme; pass it to POPPopupBridge(...). App switch disabled.")
```
jaxdesmarais: This is not a pattern we should use and illustrates the earlier comment around needing a second init with `returnURLScheme` required. The default init should not change.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -105,20 +213,38 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            && (appSwitchHandler?.isPayPalAppInstalled() ?? false)
+
         let javascript = PopupBridgeUserScript(
             scheme: PopupBridgeConstants.callbackURLScheme,
             scriptMessageHandlerName: messageHandlerName,
-            host: hostName,
-            isVenmoInstalled: application.isVenmoAppInstalled()
+            host: PopupBridgeConstants.host,
+            isVenmoInstalled: application.isVenmoAppInstalled(),
+            isPayPalInstalled: isPayPalAppSwitchAvailable,
```
jaxdesmarais, line 228: Once we clean up the return URL we can instead just check `enablePayPalAppSwitch` and `isPayPalAppInstalled` here.

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -129,42 +255,64 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            injectWebView(webView: webView, withJavaScript: cancelOrCompleteScript)
+            return
+        }
+
+        startWebAuthenticationSession(url: url)
+    }
+
+    /// Starts an `ASWebAuthenticationSession` for the URL, forwarding the return URL on success and
+    /// cancelling the flow if the user dismisses the session.
+    private func startWebAuthenticationSession(url: URL) {
```
```suggestion
    private func startWebAuthenticationSession(with url: URL) {
```

Sources/PopupBridge/PayPalAppSwitchHandler.swift
```diff
@@ -0,0 +1,173 @@
...
+    // MARK: - Return URL Scheme
+
+    /// Resolves the URL scheme used as the checkout return URL.
+    ///
+    /// The app switch path requires an explicit `returnURLScheme`: guessing the first scheme from
+    /// `CFBundleURLTypes` is fragile for apps that register multiple schemes (e.g. Facebook, Google
+    /// Sign-In), so PopupBridge fails fast instead of guessing. When `enablePayPalAppSwitch` is
+    /// `true`, `POPPopupBridge` requires this to be non-`nil`.
+    /// - Returns: The explicitly provided return URL scheme, or `nil` if none was given.
+    private func resolveReturnURLScheme() -> String? {
```
jaxdesmarais: We can clean this up once return URL is required for this flow

Sources/PopupBridge/POPPopupBridge.swift
joaaraujo-paypal: I think it doesn't quite work here, I changed the `aplication` to be a `var` with a default value. But it's consumed inside `init` (it builds the appSwitchHandler and the injected user script). Setting it via property after init is too late, the `isPayPalInstalled` user-script tests check what's generated at init time, so they need the mock injected through the initializer.
The property is private, and @testable only exposes internal. To set it directly from tests I'd have to drop private and expose mutable internal state, which is less encapsulated than a test-only init.
So I think keep these convenience inits is better, they're the controlled injection points that let both stay private.
What do you think?

Sources/PopupBridge/POPPopupBridge.swift
joaaraujo-paypal: Same here from the previously answer, we need it for tests.

Sources/PopupBridge/POPPopupBridge.swift
joaaraujo-paypal: I just move it to `URLDetailsPayload`, because now we use it on other files, like PayPalAppSwitch,

jaxdesmarais (commented):
We should just need these inits. I didn't run through the flows but updated to this on your fork and tests pass locally. This still allows us to keep application a var on the class with a default value:
```swift
    public convenience init(
        webView: WKWebView,
        returnURLScheme: String,
        prefersEphemeralWebBrowserSession: Bool = true
    ) {
        self.init(
            webView: webView,
            returnURLScheme: returnURLScheme,
            prefersEphemeralWebBrowserSession: prefersEphemeralWebBrowserSession,
            application: UIApplication.shared,
            webAuthenticationSession: WebAuthenticationSession()
        )
    }

    public convenience init(
        webView: WKWebView,
        prefersEphemeralWebBrowserSession: Bool = true
    ) {
        self.init(
            webView: webView,
            returnURLScheme: nil,
            prefersEphemeralWebBrowserSession: prefersEphemeralWebBrowserSession,
            application: UIApplication.shared,
            webAuthenticationSession: WebAuthenticationSession()
        )
    }

    /// Exposed for testing. 
    convenience init(
        webView: WKWebView,
        webAuthenticationSession: WebAuthenticationSession,
        returnURLScheme: String? = nil,
        application: URLOpener = UIApplication.shared
    ) {
        self.init(
            webView: webView,
            returnURLScheme: returnURLScheme,
            prefersEphemeralWebBrowserSession: true,
            application: application,
            webAuthenticationSession: webAuthenticationSession
        )
    }
```

Sources/PopupBridge/UIApplication+URLOpener.swift
```diff
@@ -3,15 +3,29 @@ import UIKit
...
+    func isPayPalAppInstalled() -> Bool {
+        guard let payPalURL = URL(string: "paypal://") else {
+            return false
+        }
+        return canOpenURL(payPalURL)
+    }
+
+    func openURL(_ url: URL, completionHandler: @escaping (Bool) -> Void) {
+        open(url, options: [:], completionHandler: completionHandler)
+    }
```
jaxdesmarais, line 30: Why do we need this now for PP but not before?

Sources/PopupBridge/PopupBridgeConstants.swift
```diff
@@ -1,6 +1,8 @@
 import Foundation
 
-class PopupBridgeConstants {
+enum PopupBridgeConstants {
```
jaxdesmarais: why this change?

Sources/PopupBridge/UIApplication+URLOpener.swift
joaaraujo-paypal: We need it because now it launches an external app. Before, PopupBridge only opened the popup in an ASWeb, so it never needed to launch a native app. The PayPal app switch flow does: PayPalAppSwitchHandler.launch(_:) calls application.openURL(url) to open the native PayPal app and uses the success result to either register as the pending handler or fall back. It lives on the URLOpener protocol so we can mock it in tests and cover both the success and failure paths.

Sources/PopupBridge/PopupBridgeConstants.swift
joaaraujo-paypal: I changed it to an `enum` because this type only holds `static let` constants, a class has an implicit init, so it can be instantiated like `PopupBridgeConstants()`, which doesn't make sense for a namespace of constants. That said, this is unrelated to the scope of this PR, so I'll revert it back to a class and keep my change here limited to adding the new constant.

joaaraujo-paypal:
> We should just need these inits. I didn't run through the flows but updated to this on your fork and tests pass locally. This still allows us to keep application a var on the class with a default value:
> 
> ```swift
>     public convenience init(
>         webView: WKWebView,
>         returnURLScheme: String,
>         prefersEphemeralWebBrowserSession: Bool = true
>     ) {
>         self.init(
>             webView: webView,
>             returnURLScheme: returnURLScheme,
>             prefersEphemeralWebBrowserSession: prefersEphemeralWebBrowserSession,
>             application: UIApplication.shared,
>             webAuthenticationSession: WebAuthenticationSession()
>         )
>     }
> 
>     public convenience init(
>         webView: WKWebView,
>         prefersEphemeralWebBrowserSession: Bool = true
>     ) {
>         self.init(
>             webView: webView,
>             returnURLScheme: nil,
>             prefersEphemeralWebBrowserSession: prefersEphemeralWebBrowserSession,
>             application: UIApplication.shared,
>             webAuthenticationSession: WebAuthenticationSession()
>         )
>     }
> 
>     /// Exposed for testing. 
>     convenience init(
>         webView: WKWebView,
>         webAuthenticationSession: WebAuthenticationSession,
>         returnURLScheme: String? = nil,
>         application: URLOpener = UIApplication.shared
>     ) {
>         self.init(
>             webView: webView,
>             returnURLScheme: returnURLScheme,
>             prefersEphemeralWebBrowserSession: true,
>             application: application,
>             webAuthenticationSession: webAuthenticationSession
>         )
>     }
> ```

Oh nice, thanks! I didn't fully get what you meant before. I've changed it now, it reads much better.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # PopupBridge iOS Release Notes
 
+## unreleased
+
+* Add support for the PayPal app switch flow. When enabled, checkout launches the native PayPal app and falls back to `ASWebAuthenticationSession` if the launch fails. Requires additional `SceneDelegate` integration — see the README.
+* Add `enablePayPalAppSwitch` and `returnURLScheme` parameters to the `POPPopupBridge` initializer.
```
jaxdesmarais: Nit: I think we removed `enablePayPalAppSwitch`

buzzamus:
@joaaraujo-paypal  Do you have a video of the PopupBridge app switch functionality that was introduced in this PR? I am currently having troubles getting the Demo app to load on a real device on my work computer, so wondering if you were able to share a video of the end to end flow?

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -7,36 +7,133 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+            application: application,
+            webAuthenticationSession: webAuthenticationSession
+        )
+    }
+
+    /// Designated initializer. A non-`nil` `returnURLScheme` selects the PayPal app switch flow; when it
+    /// is `nil`, the standard browser-based flow is used and the native app switch path stays disabled.
+    /// Private so the only public entry points are the two flow-specific initializers above, keeping the
+    /// internal `URLOpener` out of the public API.
+    private init(
```
jaxdesmarais: iirc we don't need this one. The convenience init above is internal.

Sources/PopupBridge/PopupBridgeConstants.swift
```diff
@@ -3,4 +3,6 @@ import Foundation
 class PopupBridgeConstants {
 
     static let callbackURLScheme: String = "sdk.ios.popup-bridge"
+    static let host: String = "popupbridgev1"
+
```
jaxdesmarais: nit: remove extra linebreak

Sources/PopupBridge/PayPalAppSwitchHandler.swift
```diff
@@ -0,0 +1,155 @@
...
+
+    // MARK: - Coordinator Queries
+
+    func isPayPalAppInstalled() -> Bool {
+        application.isPayPalAppInstalled()
+    }
+
+    // MARK: - Launch
+
+    /// Attempts a native PayPal app switch for the given URL. On success, registers as the pending
```
agedd: nit: can we newline the `onSuccess` and `onFailure` doc strings

joaaraujo-paypal:
> @joaaraujo-paypal Do you have a video of the PopupBridge app switch functionality that was introduced in this PR? I am currently having troubles getting the Demo app to load on a real device on my work computer, so wondering if you were able to share a video of the end to end flow?

@buzzamus 
Here we have a page with some demos
https://paypal.atlassian.net/wiki/spaces/MI/pages/2878756128/PopupBridge+iOS+Demos

README.md
```diff
@@ -83,6 +83,30 @@ Quick Start
...
+
+    class SceneDelegate: UIResponder, UIWindowSceneDelegate {
+
+        // ... your existing SceneDelegate code ...
+
+        func scene(_ scene: UIScene, openURLContexts URLContexts: Set<UIOpenURLContext>) {
+            guard let url = URLContexts.first?.url else { return }
+            PopupBridgeAppContextSwitcher.shared.handleReturnURL(url)
+        }
+    }
```
jaxdesmarais, line 136: Few cleanup changes:
- Add scene delegate change to demo app
- Add allowlisting change to `Info.plist` 
- Add README section same as venmo for `Info.plist` allowlisting for app install checks

If you'd prefer for that to be a fast follow, that is okay with me too. You will also want to update the underlying demo app repo to point to these changes so paypal app switch can be tested as part of our demo.

joaaraujo-paypal: I did all these changes to this repo, do you know how can I change the demo page (https://braintree.github.io/popup-bridge-example/) ?

jaxdesmarais: Great question, that codebase is here: https://github.com/braintree/popup-bridge-example. Feel free to open a PR to it as well bumping the versions to one that supports this. Do you know if this only works for the PayPal SDK directly? If so which versions? Do you all plan to add support to the Braintree web SDK for app switch via popup bridge? This should all be called out in the README.

jaxdesmarais (approved):
Thanks for working through this with us. Changes look good from my end and this comment in non blocking but would be good to clarify if there are specific requirements: https://github.com/braintree/popup-bridge-ios/pull/95#discussion_r3453235708.

Approving and can merge once @buzzamus or @agedd signs off as well.

buzzamus (approved):

joaaraujo-paypal:
We did some tests again, and now we can merge it.

agedd (approved):

PR #99, opened 2026-06-22
Fix Venmo App Switch

jaxdesmarais:
### Summary of changes
  - Add `POPPopupBridge(webView:returnURLScheme:prefersEphemeralWebBrowserSession:)` initializer to support the Venmo app switch flow.
  - Use a single **effective return URL scheme** consistently across the injected JS `getReturnUrlPrefix()`, the `ASWebAuthenticationSession` `callbackURLScheme`, and the return-URL validation. The merchant scheme is used when a `returnURLScheme` is provided **and** the Venmo app is installed; otherwise the SDK's internal `sdk.ios.popup-bridge` scheme. This keeps Venmo app switch working while fixing the ASWeb popup return (previously only `getReturnUrlPrefix()` switched, leaving the callback scheme/validation mismatched → endless spinner).
  - Make `URLOpener` injectable and funnel initializers through one internal designated init (keeps `URLOpener` out of the public API and enables testing Venmo-installed behavior).
  - Add a configurable `callbackURLScheme` to `WebAuthenticationSession`.
  - Demo app: register the return URL scheme under `CFBundleURLTypes` in `Info.plist` and use the new initializer.
  - Add unit tests + `MockURLOpener`; update README and CHANGELOG.

  ### AI Usage

  **Which AI Agent Was Used?**
  - [ ] Copilot
  - [x] Claude
  - [ ] Other (Type Name Here)

  **How was AI used?**
Design discussion and trade-off analysis (single-scheme vs. two-prefix vs. one-effective-scheme), code generation for the initializer refactor and scheme-handling logic, debugging the ASWeb return spinner (root-caused to the callback-scheme/validation mismatch), writing unit tests, and updating docs.

  **Estimated AI Code Contribution**
  - [ ] less than 30%
  - [ ] 30 - 60%
  - [x] 60 - 100%

  ### Checklist
  - [x] Added a changelog entry
  - [x] Tested and confirmed payment flows affected by this change are functioning as expected

### Testing
I tested the Venmo via BT flow and confirmed Venmo app switch now works as expected. Additionally, the PayPal SDK (ASWeb) flow still works as expected. The PayPal via BT flow currently does not work as there is an outage, but clicking the X on the ASWeb flow produces the expected cancel result returned, which confirms it should also work. Please test all flows as well to confirm (you likely need to turn off VPN + quit Xcode before you can run on device).

  ### Authors
  - @jaxdesmarais


Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -16,27 +16,91 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
     private let hostName = "popupbridgev1"
     private let sessionID = UUID().uuidString.replacingOccurrences(of: "-", with: "")
     private let webView: WKWebView
-    private let application: URLOpener = UIApplication.shared
+    private let returnURLScheme: String?
+    private var application: URLOpener = UIApplication.shared
```
buzzamus, line 21: This is probably something obvious i'm overlooking, but could this be `private let application` like it was originally? I don't see it changed anywhere

buzzamus (approved):
Just one clarifying question, but other than that this looks good

Sources/PopupBridge/POPPopupBridge.swift
jaxdesmarais: We need it to be a var to set in the internal init here for unit tests:
```swift
init(
        webView: WKWebView,
        webAuthenticationSession: WebAuthenticationSession,
        returnURLScheme: String? = nil,
        application: URLOpener = UIApplication.shared,
        prefersEphemeralWebBrowserSession: Bool = true
    )
```

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -16,27 +16,91 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
     
     private var webAuthenticationSession: WebAuthenticationSession = WebAuthenticationSession()
     private var returnBlock: ((URL) -> Void)? = nil
-    
+
+    /// The URL scheme used consistently across the return URL prefix advertised to the JS
+    /// (`getReturnUrlPrefix()`), the ASWeb `callbackURLScheme`, and the return-URL validation. When a
+    /// `returnURLScheme` was provided and the Venmo app is installed, the flow app switches into the
```
agedd: small nit: 
```suggestion
    /// `returnURLScheme` is provided and the Venmo app is installed, the flow app switches into the
```

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -16,27 +16,91 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
+        webAuthenticationSession: WebAuthenticationSession,
+        returnURLScheme: String? = nil,
+        application: URLOpener = UIApplication.shared,
+        prefersEphemeralWebBrowserSession: Bool = true
+    ) {
         self.webView = webView
-        
+        self.returnURLScheme = returnURLScheme
+        self.application = application
+        self.webAuthenticationSession = webAuthenticationSession
```
agedd: small nit (take it or leave it): can we move the `self.webAuthenticationSession = webAuthenticationSession` right after `webView` that way we match the parameter order

agedd (approved):
very small nits, this looks great to me! 🚀

Sources/PopupBridge/POPPopupBridge.swift
```diff
@@ -16,27 +16,91 @@ public class POPPopupBridge: NSObject, WKScriptMessageHandler {
...
```
```suggestion
        self.webAuthenticationSession = webAuthenticationSession
        self.returnURLScheme = returnURLScheme
        self.application = application
```

Sources/PopupBridge/POPPopupBridge.swift
jaxdesmarais: done: https://github.com/braintree/popup-bridge-ios/pull/99/commits/29d52b4d5c106b927c640bc1eabe8a4953e3c45e

